PR #9, opened 2021-05-26
TASK-7294: TECH_US06: Create Notes rest service

mkrout:


wiki-service/src/main/java/org/exoplatform/wiki/service/rest/NotesRestService.java
```diff
@@ -99,28 +100,8 @@ public NotesRestService(WikiService wikiService, ResourceBundleService resourceB
...
-      return Response.ok(outputText, MediaType.TEXT_HTML).cacheControl(cc).build();
-    } catch (Exception e) {
-      log.error("Error while converting wiki page content: " + e.getMessage(), e);
-      return Response.serverError().entity(e.getMessage()).cacheControl(cc).build();
-    }
-  }
-
   @GET
-  @Path("/{wikiType}/{wikiOwner:.+}/pages/{pageId}")
+  @Path("/pages/{wikiType}/{wikiOwner:.+}/{pageId}")
```
plamarque: should be /notes

mkrout: /notes is the path of the nodes's rest service here we added /pages referring to the data type (we may have other calls for other data types as comments, attachments, wikis...)
So in this case the rest call of this method is ../rest/nodes/pages/{wikiType}/{wikiOwner:.+}/{pageId}

plamarque: then I would prefer /notes/note/

page was the name for wikis. we changed the terminology. Now it’s notebooks for wikis and notes for page

mkrout: Ok, I'll change it

PR #26, opened 2021-06-03
TASK-7294: Update Notes rest service: add delete/move and new rest AP…

mkrout:
…I by note Id

### DIFF
--- a/wiki-service/src/main/java/org/exoplatform/wiki/service/rest/NotesRestService.java
+++ b/wiki-service/src/main/java/org/exoplatform/wiki/service/rest/NotesRestService.java
@@ -24,6 +24,8 @@ import javax.servlet.http.HttpServletRequest;
 import javax.ws.rs.*;
 import javax.ws.rs.core.*;
 
+import io.swagger.annotations.*;
+import io.swagger.jaxrs.PATCH;
 import org.apache.commons.collections.CollectionUtils;
 import org.apache.commons.lang.StringUtils;
 
@@ -49,20 +51,18 @@ import org.exoplatform.wiki.tree.utils.TreeUtils;
 import org.exoplatform.wiki.utils.Utils;
 import org.exoplatform.wiki.utils.WikiHTMLSanitizer;
 
-import io.swagger.annotations.ApiParam;
 
-/**
- * Notes REST service
- */
-@SuppressWarnings("deprecation")
 @Path("/notes")
+@Api(value = "/notes", description = "Managing notes")
+@RolesAllowed("users")
+
 public class NotesRestService implements ResourceContainer {
 
   private final WikiService           noteBookService;
 
   private final ResourceBundleService resourceBundleService;
 
-  private static Log                  log = ExoLogger.getLogger("wiki:WikiRestService");
+  private static Log                  log = ExoLogger.getLogger(NotesRestService.class);
 
   private final CacheControl          cc;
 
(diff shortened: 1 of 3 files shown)


lbengayed (approved):

wiki-service/src/main/java/org/exoplatform/wiki/service/rest/NotesRestService.java
```diff
@@ -78,10 +78,13 @@ public NotesRestService(WikiService noteBookService, ResourceBundleService resou
...
-                          @PathParam("noteId") String noteId) {
+  @ApiOperation(value = "Get note by notes params", httpMethod = "GET", response = Response.class, notes = "This get the not if the authenticated user has permissions to view the objects linked to this note.")
+  @ApiResponses(value = { @ApiResponse(code = 200, message = "Request fulfilled"),
+          @ApiResponse(code = 400, message = "Invalid query input"), @ApiResponse(code = 403, message = "Unauthorized operation"),
+          @ApiResponse(code = 404, message = "Resource not found") })
+  public Response getNote(@ApiParam(value = "NoteBook Type", required = true) @PathParam("noteBookType") String noteBookType,
+                          @ApiParam(value = "NoteBook Owner", required = true) @PathParam("noteBookOwner") String noteBookOwner,
+                          @ApiParam(value = "Note id", required = true) @PathParam("noteId") String noteId) {
     try {
       Page note = noteBookService.getPageOfWikiByName(noteBookType, noteBookOwner, noteId);
```
plamarque, line 89: no permission check? how do we ensure only authorised readers will be able to load a page?
For example in getNoteById(), you have such a permission check 

`
Identity identity = ConversationState.getCurrent().getIdentity();      
Page note = noteBookService.getPageById(noteId);      
if (!noteBookService.hasPermissionOnPage(note, PermissionType.VIEWPAGE, identity)) {        return Response.status(Response.Status.FORBIDDEN).build();      }`


why not here?

wiki-service/src/main/java/org/exoplatform/wiki/service/rest/NotesRestService.java
```diff
@@ -91,30 +94,68 @@ public Response getNote(@Context UriInfo uriInfo,
...
+          return Response.status(Response.Status.BAD_REQUEST).build();
+        }
+      }
+      if(StringUtils.isEmpty(noteBookType) || StringUtils.isEmpty(noteBookOwner)){
+        return Response.status(Response.Status.BAD_REQUEST).build();
+      }
       /* TODO: check noteBook permissions */
-      Wiki noteBook = noteBookService.getWikiByTypeAndOwner(note.getWikiType(), note.getWikiOwner());
+      Wiki noteBook = noteBookService.getWikiByTypeAndOwner(noteBookType, noteBookOwner);
       if (noteBook == null) {
```
plamarque, line 159: @mkrout shouldn’t we check that user has permission to write in this nodebook before creating the page?

mkrout: In this case the check of permission is already done in the service layer (old wiki implementation): noteBookService.getWikiByTypeAndOwner() return the note only if the current user has the view permission (https://github.com/Meeds-io/notes/blob/develop/wiki-service/src/main/java/org/exoplatform/wiki/service/impl/WikiServiceImpl.java#L488) so even if the page exists if the user don't have the view permission the method will return null.

plamarque: I see but I find it weid that some endpoints take tha responsability tu permission check and others no. 
Don’t we have a guideline to always start with permission checks at the endpoint layer?

mkrout: You're right, and that what I did in all other calls using getPagebyId() since the permission check is not done on the service layer. for the calls using getWikiByTypeAndOwner() as I said the check is done in the service and it's not recommended but that's how the old implementation was done and we can't change it now since it's used in the old wiki application, We will change that when we remove the old portlet.

plamarque: What problem is there in having permission check here and duplicated in the service layer? This way, you don’t leave any debt behind in case we’d decide to build a new service layer.

PR #30, opened 2021-06-08
task-44484: NoteEdit_US01: Edit Note

lbengayed:


wiki-webapp/src/main/webapp/skin/less/notes/notes.less
```diff
@@ -49,11 +49,9 @@
 }
 
 #notesEditor{
-	.formInputGroup{
-		#cke_1_contents{
-			height: 400px!important;
-		}
-	}
+	position: fixed;
+	width: 100%;
```
SaraBoutej, line 53: For this property you can use css class helper `width-full`

mkrout (approved):
LGTM

PR #54, opened 2021-06-17
Backend refactoring

mkrout:
According to the last best practices recommendation from Archi Guild, the Note management service should be updated to add the permissions check on the service layer instead of the rest layer. To do that a new Notes service class was added because we can't make these changes on the old service since it's still used by the WebUI portlet.
Also, JavaDoc was reviewed and updated to fix the build on all needed classes.


lbengayed (dismissed):

wiki-service/src/test/java/org/exoplatform/wiki/service/TestNoteService.java
```diff
@@ -0,0 +1,217 @@
...
+  public void testGetGroupPageById() throws WikiException {
+    Wiki wiki = getOrCreateWiki(wService, PortalConfig.GROUP_TYPE, "/platform/users");
+    Identity root = new Identity("root");
+
+    assertNotNull(noteService.getNoteOfNoteBookByName(PortalConfig.GROUP_TYPE, "platform/users", "WikiHome")) ;
+
+    try {
+      noteService.createNote(wiki, "WikiHome", new Page("testGetGroupPageById-101", "testGetGroupPageById-101"),root);
+    } catch (IllegalAccessException e) {
+      e.printStackTrace();
```
plamarque: ? why not letting the test fail?

plamarque (commented):
Looks like a huge refactoring. It's hard to tell what was the intent. Could you explain why we needed this refactoring in the pr description?

wiki-service/src/test/java/org/exoplatform/wiki/service/TestNoteService.java
mkrout: Because the exception here is related to the note creation, not to the get of the note which is the purpose of the test case, But I think that you are right it will better to fail if the note is not created, I'll update it

lbengayed (approved):

PR #67, opened 2021-06-23
Task-45862 fix when delete page it’s not removed from treeview drawer

SaraBoutej:
When I delete a note page, it should be removed from the review drawer!!


mkrout (approved):
LGTM

PR #70, opened 2021-06-23
Task 45861 45889 develop

SaraBoutej:


plamarque (commented):
is it really enough to hide the button on frontend only? Shouldn’t we prevent this change on the service later in backend too?

PR #88, opened 2021-07-05
Task-46466-45965:Add + button for Insert options and Display Only functional features in Insert Options drawer

lbengayed:


mkrout (dismissed):
LGTM

wiki-webapp/src/main/webapp/javascript/eXo/wiki/ckeditor/plugins/insertOptions/plugin.js
```diff
@@ -0,0 +1,36 @@
+/**
+ * Copyright (c) 2014-2021, CKSource - Frederico Knabben. All rights reserved.
```
plamarque: if this is some eXo code, copyright should go to eXo and license be LGPL

SaraBoutej (approved):

PR #147, opened 2021-07-29
Task-45373: Fix Notes application name and description are not displayed by their value in space settings interface (#144)

NaderSai:
Problem: the Notes application name and description are not displayed by their values in space settings interface.
How it was solved: by adding the Portlets_en.properties file which will configure the default values of name and description of Notes app.


ahamdi:
Description is required

PR #158, opened 2021-08-05
Task-47516 : For simple user, I can't add note

Rdidmohamedelhadi:
change the permission of notes-editor & clean code


mkrout (approved):
LGTM

PR #167, opened 2021-08-09
TASK-43255 : fixed notes unsearchable from unified search

GouadriaHanen:
with unifiedSearch only we get the keyword and the type while searching wiki needs keyword a wikitype and wikiOwner
fixed by adding a case when wikitype and wikiOwner are null

### DIFF
--- a/notes-service/src/main/java/org/exoplatform/wiki/jpa/JPADataStorage.java
+++ b/notes-service/src/main/java/org/exoplatform/wiki/jpa/JPADataStorage.java
@@ -30,6 +30,7 @@ import org.apache.commons.lang.StringUtils;
 import org.exoplatform.commons.api.persistence.DataInitializer;
 import org.exoplatform.commons.api.persistence.ExoTransactional;
 import org.exoplatform.commons.file.services.FileService;
+import org.exoplatform.commons.utils.ListAccess;
 import org.exoplatform.commons.utils.ObjectPageList;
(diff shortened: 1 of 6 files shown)


hbenali:
@GouadriaHanen  Feel free to reopen this PR or create another once you finish working on this Task. Thanks

notes-service/src/main/java/org/exoplatform/wiki/jpa/JPADataStorage.java
```diff
@@ -38,8 +39,14 @@
 import org.exoplatform.container.xml.ValuesParam;
 import org.exoplatform.portal.config.*;
 import org.exoplatform.portal.config.model.PortalConfig;
+import org.exoplatform.services.security.ConversationState;
```
rdenarie, line 42: As explained in https://docs.google.com/spreadsheets/d/1s-op62uot1dy4YwMA-JswxNg0DwS4sdmg1eAll0VHIU/edit#gid=0, the rule DEVSTD-28 says : 

> Avoid using org.exoplatform.services.security.ConversationState.getCurrent() in Service & Storage layers. In fact, the access to the currently connected user should be done from higher layer such as REST Service / Portlet. 

So we should no get user ID in this layer.

rdenarie (changes requested):
@GouadriaHanen Can you contact me when your are available to speak about this PR.
I'm not sure to understand what you try to do.
Thanks

PR #177, opened 2021-08-25
TASK-48123: Display pasted styled tables in notes

aelbenney:
The problem was in the WikiHTMLSanitizer when the notes body contains a comment and just below a closing tag. After splitting the body, the second part will contain closing tags without their opening ones. So in the HTMLSanitizer.sanitize() method, they will be removed.
This process was done with the old wiki app to conserve the xwiki content in comments. We no longer need it in the new notes app.

This fix should replace the `markupSanitize` with the `HTMLSanitizer` to remove useless html comments.


rdenarie (changes requested):
Pr build fail

rdenarie:
@ahamdi can you take a look on this PR.
Wdyt about using HTMLSanitizer instead of WikiHTMLSanitizer ?

ahamdi:
@rdenarie WikiHTMLSanitizer was introduced tgo avoid losing Xwiki synta that was added in comments inside the source of the wiki pages. It is no more needed with the reworked Notes application

ahamdi (dismissed):

rdenarie (changes requested):
Can you rebase your pr to have only your commit ?

rdenarie (approved):

PR #178, opened 2021-08-26
Task-47516:For simple user, I can't add note

lbengayed:
After change the access permissions for notes editor page from spaces/exo_employees to /platform/users we should change the importmode to merge in order to merge the new permissions


rdenarie (approved):

rdenarie (changes requested):
Can you rename the PR to use the original task which is 47516.
Can you also rename the commit message.

rdenarie (approved):

PR #179, opened 2021-08-30
Task-48426: Missing label for the title of note editor page

AzmiTouil:
Prio this change , a technical label on the title of note editor page is displayed
Fix : move translation label  to navigation.portal.global.properties file.

### DIFF
--- a/notes-webapp/src/main/resources/locale/navigation/portal/global_en.properties
+++ b/notes-webapp/src/main/resources/locale/navigation/portal/global_en.properties
@@ -17,3 +17,6 @@
 # 02110-1301 USA, or see the FSF site: http://www.fsf.org.
 #
 portal.global.wiki=Notes
+portal.global.notes-editor= Notes Editor
+portal.global.notes=Notes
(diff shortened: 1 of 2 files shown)


notes-webapp/src/main/resources/locale/navigation/portal/global_en.properties
```diff
@@ -17,3 +17,6 @@
 # 02110-1301 USA, or see the FSF site: http://www.fsf.org.
 #
 portal.global.wiki=Notes
+portal.global.notes-editor= Notes Editor
```
rdenarie:
```suggestion
portal.global.notes-editor=Notes Editor
```

rdenarie (changes requested):
Can you update your commit message ?

You answer the question WHY ? but not the question HOW ?

rdenarie (changes requested):
You update only the PR description.
Can you update also the commit message ?
Thanks

rdenarie (approved):

PR #184, opened 2021-09-01
MAINT-48425: Allow open inserted links via the notes editor in a new tab

hakermi:
**ISSUE**: When inserting a an external link via the note editor inside the page, the insert plugin doesn't add target_blank to the inserted html "a" element.
**FIX**: Add a format function to add the target _blank attribute inside the external links in the notes content


ahamdi (changes requested):
Looping on all links is not a good solution, it may cause a performance issue when opening pages containing plenty of links.
The attribute target _blank should be added when inserting the link.
If it is external, we open it in new Note page, if it is internal we open it in the same page.

hakermi:
> Looping on all links is not a good solution, it may cause a performance issue when opening pages containing plenty of links.
> The attribute target _blank should be added when inserting the link.
> If it is external, we open it in new Note page, if it is internal we open it in the same page.

I close this PR as i added the fix in this PR : https://github.com/Meeds-io/commons/pull/238

PR #222, opened 2021-09-15
Task-48721: Image disappear from Notes

Rdidmohamedelhadi:
when we add an image in the note this image disappear after 24H, the cause of this problem during the creation of note the values of wikiType and wikiOwner are Null, my fix is to pass the values of wikiOwner and wikiType in url and use it when creating.


AzmiTouil (changes requested):
Fix Sonar issue please!

Rdidmohamedelhadi:
> Fix Sonar issue please!

done !

AzmiTouil (approved):

PR #249, opened 2021-09-28
Task-46383: Space applications name translation is missing

AzmiTouil:
Prior this fix, the space application name display as a technical labels

### DIFF
--- a/notes-webapp/src/main/resources/locale/portlet/Portlets_en.properties
+++ b/notes-webapp/src/main/resources/locale/portlet/Portlets_en.properties
@@ -1,0 +1,9 @@
+#############################################################################
+#SpaceApplications
+#############################################################################
+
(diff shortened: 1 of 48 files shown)


ahamdi (approved):

notes-webapp/src/main/resources/locale/portlet/Portlets_en.properties
```diff
@@ -0,0 +1,9 @@
+#############################################################################
+#SpaceApplications
+#############################################################################
+
+SpaceSettings.application.wiki.title=Notes
+SpaceSettings.application.wiki.description=Wiki Portlet
```
plamarque, line 6: The old wiki app is supposed to have been removed why are we keeping this ? 

CC @mkrout

PR #250, opened 2021-09-28
Task-49207: Export_note_US_01-03: Export notes

lbengayed:


notes-service/src/main/java/org/exoplatform/wiki/service/NoteService.java
```diff
@@ -167,7 +167,7 @@ public boolean renameNote(String noteType,
    * @return The list of children notes
    * @throws WikiException if an error occured
    */
-  List<Page> getChildrenNoteOf(Page note, boolean withDrafts) throws WikiException;
+  public List<Page> getChildrenNoteOf(Page note, boolean withDrafts) throws WikiException;
```
mdiha: Since its an interface no need to put the keyword "public" on method signature.

mkrout (approved):

PR #273, opened 2021-10-13
Task-49978: Filtre_notes_US_02 : When there is no result for search

lbengayed:


notes-webapp/src/main/webapp/vue-app/notes/components/NoteTreeviewDrawer.vue
```diff
@@ -88,7 +89,7 @@
               </div>
             </v-col>
           </v-row>
-          <template v-if="home && !exportNotes && filter !== $t('notes.filter.label.drafts')" class="ma-0 border-box-sizing">
+          <template v-if="home && !exportNotes && showTree && filter !== $t('notes.filter.label.drafts')" class="ma-0 border-box-sizing">
```
mdiha: It is better to set showTree as a computed prop, to avoid handling its value everywhere

notes-webapp/src/main/webapp/vue-app/notes/components/NoteTreeviewDrawer.vue
```diff
@@ -201,12 +213,13 @@ export default {
     filter: '',
     filterOptions: [],
     keyword: '',
+    active: [],
     checkbox: false,
+    showTree: true,
+    search: null,
```
mdiha: search is a string, better set its default value to empty string ''

notes-webapp/src/main/webapp/vue-app/notes/components/NoteTreeviewDrawer.vue
```diff
@@ -425,6 +441,7 @@ export default {
       this.$refs.breadcrumbDrawer.close();
     },
     closeAllDrawer() {
+      this.search = null;
```
mdiha: search is a string, better set its default value to empty string ''

mdiha (approved):
LGTM

PR #332, opened 2021-10-28
Backport-Task-47335: Add target blank to external links

SaraBoutej:
Before this fix when we click on external link in the note page, it always opened in the same tab.
This PR add target= _blank when there is an external lint, to be opened in a new tab


hbenali (commented):
@SaraBoutej Do we still need to backport this PR?

PR #336, opened 2021-10-28
Task-49216: Add role presentation for table tag

SaraBoutej:
Before this fix all tables inserted in note editor don't have **role="presentation"**
In this Commit we add **role="presentation"** to improve accessibility in the platform


mdiha (approved):
LGTM

PR #390, opened 2021-11-29
Task-51804 : Error during startup

rdenarie:
Before this fix, there an error in server log.

This error is due to the usage of ConversationState in NotesAutoImportService.
This commit removes the usage of the current conversation state.


mkrout (approved):
LGTM

mkrout (approved):

PR #428, opened 2021-12-23
Task-52445: Re-implement exportNotes method in order to add the ability of exporting independent notes.

mdiha:
Currently the export notes functionality doesn't allow to export notes independently which means that when exporting a note, its parent is automatically selected in the notes to export, and so it is not possible to export a parent note independently since when selecting a note all its children are selected.
These changes add the ability to export notes independently which allows also to restructure the tree note to export according to selection.
For example: we have tree note: 
home > Note01 > Note01.1 > Note01.1.1
                          > Note01.2
          > Note02 > Note02.1
                          > Note02.2
when selecting notes (Note01, Note01.1.1, Note02.2) to be exported
the result tree would be:
home > Note01 > Note01.1.1
          > Note02.2
and so.

### DIFF
--- a/notes-service/src/main/java/org/exoplatform/wiki/mow/api/NoteToExport.java
+++ b/notes-service/src/main/java/org/exoplatform/wiki/mow/api/NoteToExport.java
@@ -16,14 +16,14 @@
  */
 package org.exoplatform.wiki.mow.api;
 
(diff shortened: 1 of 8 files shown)


mkrout (approved):
LGTM

PR #448, opened 2022-01-06
Task-52643: Make sure that when saving a new note the according draft note is well deleted.

mdiha:
Issue: the previously added window.location.href code line comes just after the call to the deleteDraftNote function which contains a fetch call, which makes the fetch call aborted and throws an error: "TypeError: NetworkError when attempting to fetch resource".
Reproduced only on FireFox.
Fix: I removed this line of code since it is already set in the promise's resolve of the fetch call response in the function deleteDraftNote.

### DIFF
--- a/notes-webapp/src/main/webapp/vue-app/notes-editor/components/NotesEditorDashboard.vue
+++ b/notes-webapp/src/main/webapp/vue-app/notes-editor/components/NotesEditorDashboard.vue
@@ -422,7 +422,6 @@ export default {
             // delete draft note
             const draftNote = JSON.parse(localStorage.getItem(`draftNoteId-${this.note.id}`));
             this.deleteDraftNote(draftNote, notePath);
-            window.location.href = notePath;
           }).catch(e => {
             console.error('Error when creating note page', e);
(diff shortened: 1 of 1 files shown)


notes-webapp/src/main/webapp/vue-app/notes-editor/components/NotesEditorDashboard.vue
```diff
@@ -422,7 +422,6 @@ export default {
             // delete draft note
             const draftNote = JSON.parse(localStorage.getItem(`draftNoteId-${this.note.id}`));
             this.deleteDraftNote(draftNote, notePath);
-            window.location.href = notePath;
```
mkrout, line 425: So after saving the note and deleting the draft, no redirection to the created note?

mdiha: The redirection is already set inside the deleteDraftNote(draftNote, notePath) function call.

mkrout (approved):

SaraBoutej (approved):

PR #475, opened 2022-01-25
Task-51726 : pop up error when editing the home note page

Noussairhamrit:
ISSUES :  when User clicks on the edit button of home page note without any change an error pop-up arises notes.message.Cannot read properties of null (reading 'split') .
FIX :so the cause of the error is 'spaceDisplayName' in the 'getPathByNoteOwner' method in the 'noteService.js' class then the 'spaceDisplayName' variable receives null because note.url is null . so  i set the note.url  on 'NotesEditorDashboard.vue' to send this correct url to NotesRestService.java


notes-service/src/main/java/org/exoplatform/wiki/service/rest/NotesRestService.java
```diff
@@ -539,6 +539,8 @@ public Response updateNoteById(@ApiParam(value = "Note id", required = true) @Pa
           WikiPageParams noteParams = new WikiPageParams(note_.getWikiType(), note_.getWikiOwner(), newNoteName);
           noteService.removeDraftOfNote(noteParams);
         }
+      } else if (note_.getContent().equals(note.getContent())) {
```
rdenarie: This modification will launch listeners related to edit content event, for example update the last modification date. 

If the content is not modified we should note update the note.
Can you try to find another solution ?

Thanks

rdenarie (changes requested):

rdenarie (changes requested):
I think we could have a simplier change than what you did.

What is the problem exactly ?
When you update a note, without changing title nor content, the url in note sent by updateNoteById is empty. 
In updateNoteById, we have a big if, with 3 case :  
```java
if (title changed && content changed) {
...
} else if (title changed) {
...
} else if {content changed) {
...
} else {
nothing (not exists)
} 
```
Can we create the else in this if ? and fill url someway in this else case, so that, when we change nothing, the PUT answer a note with url filled, and the error should be removed on front.

With this solution, we dont need any change in the vue file.
The only thing to do is to find how to fill note.url. To find that, you can follow the process of updating a note when changing content. This is the working case, and you should analyze the differences between the non working case.

notes-service/src/main/java/org/exoplatform/wiki/service/rest/NotesRestService.java
```diff
@@ -539,6 +539,8 @@ public Response updateNoteById(@ApiParam(value = "Note id", required = true) @Pa
           WikiPageParams noteParams = new WikiPageParams(note_.getWikiType(), note_.getWikiOwner(), newNoteName);
           noteService.removeDraftOfNote(noteParams);
         }
+      } else{
+         note_.setUrl(Utils.getPageUrl(note));
```
rdenarie:
```suggestion
      } else{
         //in this case, the note didnt change on title nor content. As we need the page url in front side, we compute it here
         note_.setUrl(Utils.getPageUrl(note));
```

rdenarie (changes requested):

rdenarie (changes requested):
you should have not modification in any pom.xml in your pr.
Only the java file

rdenarie (approved):

PR #479, opened 2022-01-31
Task-50373 : Display order of note tree is not respected

rdenarie:
Before this fix, when a page display a note tree, in some case, pages are not alphabetically ordered
This problem is due to the calcultation of the tree, which manipulate pages children without keeping the children order
This fix ensure that when we remove an item from a list, we keep the index to be able to readd it at the same place after the calculation


hakermi (approved):
LGTM

PR #485, opened 2022-02-18
TASK-50564 : paste an image copied to clipboard.

IlhemEssaadi:
ISSUES : when creating a note that contains a pasted image, the pasted image is not visible also when a draft is saved just after pasting the image then close the editor without saving, then open the saved draft, the draft do not contains the image.
FIX : the first problem is that we cannot read the new content of the ckeditor correctly because the processing of the pasted and downloaded image takes some time and it is solved by adding the getBody() function that reads the content of the ckeditor at the moment the user clicks the save button ,and the second problem is that the localStorage doesn't receive the correct value of the saved draft it is solved by adding the correct value of the new content of the saved draft.

### DIFF
--- a/notes-webapp/src/main/webapp/vue-app/notes-editor/components/NotesEditorDashboard.vue
+++ b/notes-webapp/src/main/webapp/vue-app/notes-editor/components/NotesEditorDashboard.vue
@@ -383,7 +383,7 @@ export default {
             name: this.note.name,
             wikiType: this.note.wikiType,
             wikiOwner: this.note.wikiOwner,
-            content: this.note.content,
+            content: this.getBody() ? this.getBody() : this.note.content,
             parentPageId: this.note.targetPageId === this.parentPageId ? null : this.parentPageId,
             toBePublished: toPublish,
             appName: this.appName,
@@ -395,7 +395,7 @@ export default {
             name: this.note.name,
             wikiType: this.note.wikiType,
             wikiOwner: this.note.wikiOwner,
-            content: this.note.content,
+            content: this.getBody() ? this.getBody() : this.note.content,
             parentPageId: this.parentPageId,
             toBePublished: toPublish,
             appName: this.appName,
@@ -444,7 +444,7 @@ export default {
       const draftNote = {
         id: this.note.draftPage ? this.note.id : '',
         title: this.note.title,
-        content: this.note.content,
+        content: this.getBody() ? this.getBody() : this.note.content,
         name: this.note.name,
         appName: this.appName,
         wikiType: this.note.wikiType,
@@ -461,7 +461,7 @@ export default {
         // if draft page not created persist it only the first time else update it in browser's localStorage
         if (this.note.draftPage && this.note.id) {
           this.note.parentPageId = this.parentPageId;
-          localStorage.setItem(`draftNoteId-${this.note.id}`, JSON.stringify(this.note));
+          localStorage.setItem(`draftNoteId-${this.note.id}`, JSON.stringify(draftNote));
           this.actualNote = {
             name: draftNote.name,
             title: draftNote.title,
@@ -813,7 +813,11 @@ export default {
       } else {
         return false;
       }
+    },
+    getBody: function() {
+      const newData = CKEDITOR.instances['notesContent'].getData();
+      return newData ? newData : null;
     }
-  }
+  },
 };
 </script>
--- a/notes-webapp/src/main/webapp/vue-app/notes-editor/components/NotesEditorDashboard.vue
+++ b/notes-webapp/src/main/webapp/vue-app/notes-editor/components/NotesEditorDashboard.vue
@@ -383,7 +383,7 @@ export default {
             name: this.note.name,
             wikiType: this.note.wikiType,
             wikiOwner: this.note.wikiOwner,
-            content: this.getBody() ? this.getBody() : this.note.content,
+            content: this.getBody() || this.note.content,
             parentPageId: this.note.targetPageId === this.parentPageId ? null : this.parentPageId,
             toBePublished: toPublish,
             appName: this.appName,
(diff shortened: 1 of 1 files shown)


rdenarie (changes requested):
Can you explain why we need to wait before reading the note content after a paste ?
I'm not understand currently

IlhemEssaadi:
> Can you explain why we need to wait before reading the note content after a paste ? I'm not understand currently

paste and upload image processing takes a while before reading it from the event.

rdenarie:
> > Can you explain why we need to wait before reading the note content after a paste ? I'm not understand currently
> 
> paste and upload image processing takes a while before reading it from the event.

Please add this in your description.

Did you try with a promise instead of a timeout ?

IlhemEssaadi:
> > > Can you explain why we need to wait before reading the note content after a paste ? I'm not understand currently
> > 
> > 
> > paste and upload image processing takes a while before reading it from the event.
> 
> Please add this in your description.
> 
> Did you try with a promise instead of a timeout ?

yes and does not work !!

rdenarie (changes requested):
I checked this problem I think there is a better solution than using a timeout. 

I compared with ExoNewsActivityComposer, which allow to write a news. In this component, we are able to paste an image, and publish directly. The image is correctly present in the news.
For that, this component define a function getBody() : 
`getBody: function() {
      const newData = CKEDITOR.instances['newsContent'].getData();
      return newData ? newData : null;
    },`

This function read the ckeditor content at the moment user clicks on the save button. So at this time, we are sure to have the last up-to-date content, and we can use it to save the news.

We should make the same behaviour for notes.

One remark : I found a problem in notes, when a draft is saved just after pasting the image. If you paste image, then close the editor without saving, then open the saved draft, the draft do not contains the image. The image is present if you put some characters after pasting the image.

We need to ensure that this problem is not present in noates after your fix. For the news problem we will create a new task

notes-webapp/src/main/webapp/vue-app/notes-editor/components/NotesEditorDashboard.vue
```diff
@@ -383,7 +383,7 @@ export default {
             name: this.note.name,
             wikiType: this.note.wikiType,
             wikiOwner: this.note.wikiOwner,
-            content: this.note.content,
+            content: this.getBody() ? this.getBody() : this.note.content,
```
rdenarie: here you call twice this.getBody.
Is it possible to call it once ?
you could use this.getBody() || this.note.content 
I thought it work, but please check.
Else you can use a variable

notes-webapp/src/main/webapp/vue-app/notes-editor/components/NotesEditorDashboard.vue
```diff
@@ -395,7 +395,7 @@ export default {
...
```
rdenarie: same remark

notes-webapp/src/main/webapp/vue-app/notes-editor/components/NotesEditorDashboard.vue
```diff
@@ -444,7 +444,7 @@ export default {
       const draftNote = {
         id: this.note.draftPage ? this.note.id : '',
         title: this.note.title,
-        content: this.note.content,
+        content: this.getBody() ? this.getBody() : this.note.content,
```
rdenarie: same remark

notes-webapp/src/main/webapp/vue-app/notes-editor/components/NotesEditorDashboard.vue
```diff
@@ -461,7 +461,7 @@ export default {
         // if draft page not created persist it only the first time else update it in browser's localStorage
         if (this.note.draftPage && this.note.id) {
           this.note.parentPageId = this.parentPageId;
-          localStorage.setItem(`draftNoteId-${this.note.id}`, JSON.stringify(this.note));
+          localStorage.setItem(`draftNoteId-${this.note.id}`, JSON.stringify(draftNote));
```
rdenarie, line 464: can you explain here why you change this.note to draftNote ?

notes-webapp/src/main/webapp/vue-app/notes-editor/components/NotesEditorDashboard.vue
```diff
@@ -813,7 +813,11 @@ export default {
       } else {
         return false;
       }
+    },
+    getBody: function() {
+      const newData = CKEDITOR.instances['notesContent'].getData();
+      return newData ? newData : null;
     }
-  }
+  },
```
rdenarie:
```suggestion
  }
```

rdenarie (changes requested):

notes-webapp/src/main/webapp/vue-app/notes-editor/components/NotesEditorDashboard.vue
IlhemEssaadi: > can you explain here why you change this.note to draftNote ?

this.note does not have  the last up-to-date content of ckeditor.

notes-webapp/src/main/webapp/vue-app/notes-editor/components/NotesEditorDashboard.vue
```diff
@@ -618,6 +618,7 @@ export default {
           },
           change: function (evt) {
             self.note.content = evt.editor.getData();
+            self.autoSave();
```
rdenarie, line 621: I think this is not necessary.
There is a watch on note.content, which call autosave().
Can you check

rdenarie (changes requested):

notes-webapp/src/main/webapp/vue-app/notes-editor/components/NotesEditorDashboard.vue
rdenarie: we need to keep it.

rdenarie (approved):

PR #492, opened 2022-03-21
TASK-55496 Fix Displaying Notes page children in page Overview

boubaker:
Prior to this change, a Note page isn't displayed due to the fact that HTMLCollection isn't iterable and thus we can't use forEach on it. This fix will replace forEach by a simple for loop and a test on nullability on HTMLCollection variable


mkrout (approved):
LGTM

PR #503, opened 2022-04-26
TASK-56675: Deleted notes still accessible by note page links

hakermi:
Prior to this change,when deleteing a notes its deleted flag is set to true, while when getting the note by its id not control is added to check if it's deleted or not, and so it's displayed to the user,
This PR should add a check on the dleted flag and return NOT_FOUND http state code if it's deleted


boubaker (commented):
can you please add unit tests

ahamdi (approved):

PR #505, opened 2022-04-27
TASK-56665: Notes drafts cannot be displayed in the notes tree

hakermi:
Prior to this change, when selecting drafts filter in notes tree drawer, there was an issue due to an `OutOfBoundArrayException` because a bad code placement when filter notes children depends on the `withDrafts query` param, which selecting the child index in the list of children and then filter the list and override its size while keeping and using the old index to insert the draft child in the list.
This PR should fix this problem by correctly place the code of filtering the list before selecting the child index to avoid having wrong index and an exception.


ahamdi (commented):
Could you add a UT for this ?

ahamdi (approved):

PR #506, opened 2022-04-28
Long latency when exporting notes

mkrout:


notes-service/src/main/java/org/exoplatform/wiki/service/ExportAction.java
```diff
@@ -0,0 +1,42 @@
+/*
+ * Copyright (C) 2003-2011 eXo Platform SAS.
```
plamarque, line 2: new class should be copyright 2022 meeds with LGPL

notes-service/src/main/java/org/exoplatform/wiki/service/NotesExportService.java
```diff
@@ -0,0 +1,111 @@
...
+    }
+  }
+
+  public void startExportNotes(int exportId, String[] notesToExportIds, boolean exportAll, Identity identity) throws Exception {
+    ExportResource exportResource = new ExportResource();
+    exportResource.setExportId(exportId);
+    exportResource.setStatus(ExportStatus.STARTED.name());
+    exportResource.setAction(new ExportAction());
+    exportResourceList.add(exportResource);
+    Thread exportThread = new Thread(new ExportThread(noteService,
```
plamarque, line 64: Should use an ExecutorService with fixed number of threads IMO.

plamarque (commented):
- Check new files header for copyright and license (meeds, 2022, LGPL)
- Avoid using raw threads

PR #510, opened 2022-05-04
Task-56742: fix the update for the posted notes 

sofyenne:
ISSUE : when saving & posting a note then get back to update the old title and content are displayed (the note activity not updated)
Fix : the note activity updated

### DIFF
--- a/notes-service/src/main/java/org/exoplatform/wiki/jpa/entity/PageEntity.java
+++ b/notes-service/src/main/java/org/exoplatform/wiki/jpa/entity/PageEntity.java
@@ -111,6 +111,9 @@ public class PageEntity extends BasePageEntity {
   @Column(name = "DELETED")
   private boolean deleted;
 
+  @Column(name = "TOBEPUBLISHED")
+  private boolean toBePublished;
+
   public long getId() {
     return id;
   }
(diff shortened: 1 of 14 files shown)


ahamdi:
@sofyenne Could you explain technically what was the problem and why you added that field in the database ?

notes-service/src/main/java/org/exoplatform/wiki/jpa/entity/PageEntity.java
```diff
@@ -111,6 +111,9 @@ public class PageEntity extends BasePageEntity {
   @Column(name = "DELETED")
   private boolean deleted;
 
+  @Column(name = "TOBEPUBLISHED")
+  private boolean toBePublished;
```
boubaker, line 115: This modification doesn't seem fixing the real issue and the description doesn't mention the real cause of the problem and how did you analysed it and proposed this fix. We 'll have to re-work this together.

sofyenne:
> @sofyenne Could you explain technically what was the problem and why you added that field in the database ?
@ahamdi 
If the user saved and published a note and went back to modify it, the note is updated in the note space but is not updated in the activity space .I added an attribute in the table to check the status of the note (published or not) during the update and that would solve the problem !

boubaker:
please close this since a new PR is created https://github.com/Meeds-io/notes/pull/511

PR #523, opened 2022-05-11
TASK-56521 Avoid duplicating Note Favorite with Activity

boubaker:
Prior to this change, when adding note as favorite from Activity stream, a Favorite Metadata is associated to Activity and to Note at the same time. This fix will ensure to associate the favorite Metadata to Notes page only. In fact, an improvement for Activity Metadata Management has been added that allows to add a specific target Object Type+Id when displaying activity in Stream. Thus, the Favorite button will manage the favorites of Notes Object identified by its real id.

(Remaining: **Unit Tests**)


AzmiTouil (approved):
LGTM

PR #525, opened 2022-05-12
Task-55861: Add plugin to handle file upload response

akhanfir:
Prior to this change, the image  thumbnail isn't completely displayed to end user due to missing Content-Length HTTP Header.this fix will ensure to change add method that we can get the response from server and update the editor content this method is called when file upload is finished

### DIFF
--- a/notes-webapp/src/main/webapp/vue-app/notes-editor/components/NotesEditorDashboard.vue
+++ b/notes-webapp/src/main/webapp/vue-app/notes-editor/components/NotesEditorDashboard.vue
@@ -630,6 +630,16 @@ export default {
               } );
             }
           },
+          fileUploadResponse: function() {
+            /*add plugin fileUploadResponse to handle file upload response ,
+              in this method we can get the response from server and update the editor content
+              this method is called when file upload is finished*/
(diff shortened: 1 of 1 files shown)


boubaker:
Please update title and description to reflect content of PR

notes-webapp/src/main/webapp/vue-app/notes-editor/components/NotesEditorDashboard.vue
```diff
@@ -555,7 +555,7 @@ export default {
           { name: 'links', items: [ 'simpleLink','InsertOptions'] },
         ],
         format_tags: 'p;h1;h2;h3',
-        autoGrow_minHeight: self.noteFormContentHeight,
+        autoGrow_minHeight: true,
```
boubaker: can you please check this change, in fact this attribute must be of type `Number`, see https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_config.html#cfg-autoGrow_minHeight

boubaker (approved):

boubaker:
approved, please pay attention to the messages when **Squash & Merge**

PR #529, opened 2022-05-16
Task-56989: Fix display of the informative message (#524)

lbengayed:
Prior to this change, when closing the popup information and navigating between notes, the popup keeps on being displayed.
This PR should make sure to check if the popup was unready displayed or not


mkrout (approved):
LGTM

PR #534, opened 2022-05-17
Task-55861: Add Content Length HTTP Header when reading Image Thumbnail

akhanfir:
Prior to this change, the image  thumbnail isn't completely displayed to end user due to missing Content-Length HTTP Header.this fix will ensure to change min_Height autoGrow to true and after insert image i reload the notes content by method fire


hassenelkamel-exo (approved):
LGTM

PR #537, opened 2022-05-18
TASK-56623: Notes breadcrumb doesn't show renamed home title

hakermi:
Prior to this change, note breadcumb of the the home note is overrided by an i18n label instead of displaying the renamed title.
This PR should add a function to display the renamed title or the i18n original label if the page wasn't renamed

### DIFF
--- a/notes-webapp/src/main/webapp/vue-app/notes/components/NotesOverview.vue
+++ b/notes-webapp/src/main/webapp/vue-app/notes/components/NotesOverview.vue
@@ -335,7 +335,7 @@ export default {
       }
       setTimeout(() => this.hasManualChildren = false, 100);
       if ( this.note && this.note.breadcrumb && this.note.breadcrumb.length ) {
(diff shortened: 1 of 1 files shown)


notes-webapp/src/main/webapp/vue-app/notes/components/NotesOverview.vue
```diff
@@ -506,6 +506,11 @@ export default {
     }
   },
   methods: {
+    getHomeTitle(title) {
+      const label = `notes.label.note${title}`;
+      const translation = this.$t(label);
+      return translation === label && title || translation;
```
boubaker:
```suggestion
      return title === 'Home' && this.$t('notes.label.noteHome') || title;
```
May be just test when `title === 'Home'` to see if we have to apply `this.$t` on it. This will avoid to attempt internationalizing other labels that can match an i18N key

boubaker (approved):

PR #551, opened 2022-05-26
Task-56296: unable to import note

lbengayed:
Prior to this fix, when user try to reimport note and chose the duplicate  action, An error message is displayed and pages notes are not imported “note already exists”
With this fix, we will be able to duplicate any note with change the name of page duplicated (increment the number) if it exists.

### DIFF
--- a/notes-service/src/main/java/org/exoplatform/wiki/service/impl/NoteServiceImpl.java
+++ b/notes-service/src/main/java/org/exoplatform/wiki/service/impl/NoteServiceImpl.java
@@ -1196,9 +1196,11 @@ public class NoteServiceImpl implements NoteService {
 
           }
           if (conflict.equals("duplicate")) {
-            int i = 1;
+            String title = note.getTitle();
+            int i = title.lastIndexOf("_") != -1 ? Integer.valueOf(title.substring(title.lastIndexOf("_") + 1)) + 1 : 1;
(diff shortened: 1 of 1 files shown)


notes-service/src/main/java/org/exoplatform/wiki/service/impl/NoteServiceImpl.java
```diff
@@ -1196,9 +1196,11 @@ public void importNote(Page note, Page parent, Wiki wiki, String conflict, Ident
 
           }
           if (conflict.equals("duplicate")) {
-            int i = 1;
+            String title = note.getTitle();
+            int i = title.lastIndexOf("_") != -1 ? Integer.valueOf(title.substring(title.lastIndexOf("_") + 1)) + 1 : 1;
```
mkrout: I think that we may have a NumberFormatException here since the substring after the last "_" can be not numeric. should be sure that this substring can be converted to int before incrementing it.

mkrout (changes requested):

notes-service/src/main/java/org/exoplatform/wiki/service/impl/NoteServiceImpl.java
lbengayed: done!

mkrout (approved):
LGTM

PR #567, opened 2022-06-01
Task-57855 : unable to create a wiki when using apache front

rdenarie:
Before this fix, when using apache frontend, when wikiOwner is used as pathParam, there are 2 slash in the url. The front remove one of this slash so the wiki owner is not the good one


notes-service/src/main/java/org/exoplatform/wiki/service/rest/NotesRestService.java
```diff
@@ -1312,4 +1325,17 @@ private Page updateChildrenContainer(Page note) throws WikiException {
     return noteService.updateNote(note);
   }
 
+  private String validateGroupWikiOwner(String wikiOwner) {
```
boubaker: There is not validation inside, IMO it's : 
```suggestion
  private String extractGroupWikiOwner(String wikiOwner) {
```
or
```suggestion
  private String formatWikiOwnerToGroupId(String wikiOwner) {
```

boubaker (approved):

PR #568, opened 2022-06-06
Task-54748: home label & welcome message not translated in note application

wassimmhamdi:
Problem: when opening note application the home label and the  welcome message are not translated
Fix: I tried to fix the method that return the correct translation

### DIFF
--- a/notes-webapp/src/main/webapp/vue-app/notes/components/NotesOverview.vue
+++ b/notes-webapp/src/main/webapp/vue-app/notes/components/NotesOverview.vue
@@ -336,10 +336,10 @@ export default {
       }
       setTimeout(() => this.hasManualChildren = false, 100);
       if ( this.note && this.note.breadcrumb && this.note.breadcrumb.length ) {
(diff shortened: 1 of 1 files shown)


notes-webapp/src/main/webapp/vue-app/notes/components/NotesOverview.vue
```diff
@@ -336,7 +336,7 @@ export default {
       }
       setTimeout(() => this.hasManualChildren = false, 100);
       if ( this.note && this.note.breadcrumb && this.note.breadcrumb.length ) {
-        this.note.breadcrumb[0].title = this.getHomeTitle(this.note.breadcrumb[0].title);
+        this.note.breadcrumb[0].title = this.$t('notes.label.noteHome');
```
ahamdi: You need to check what was the logic added in the function getHomeTitle() and not remove it

ahamdi (changes requested):

ahamdi (approved):

PR #569, opened 2022-06-10
Task-57821: fix uploaded images in notes appear as undefined

hchaabene:
Problem: Before this fix, only  user who added the image can see the image uploaded to the notes.
Solution: When I click on "Add Notes", I added the spaceGroupId to the URL and then restored it with urlParams.
After this change, the other users can see the uploaded image when they upload it.

### DIFF
--- a/notes-webapp/src/main/webapp/vue-app/notes-editor/components/NotesEditorDashboard.vue
+++ b/notes-webapp/src/main/webapp/vue-app/notes-editor/components/NotesEditorDashboard.vue
@@ -215,6 +215,7 @@ export default {
     if (urlParams.has('parentNoteId')) {
       this.parentPageId = urlParams.get('parentNoteId');
       this.spaceId = urlParams.get('spaceId');
+      this.spaceGroupId  = urlParams.get('spaceGroupId');
       this.note.parentPageId = this.parentPageId;
     }
     this.displayFormTitle();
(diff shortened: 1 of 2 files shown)


notes-webapp/src/main/webapp/vue-app/notes-editor/components/NotesEditorDashboard.vue
```diff
@@ -564,7 +565,7 @@ export default {
         removePlugins: removePlugins,
         allowedContent: true,
         spaceURL: self.spaceURL,
-        spaceGroupId: self.spaceGroupId,
+        spaceGroupId: `/spaces/${  this.spaceGroupId}`,
```
ahamdi, line 568:
```suggestion
        spaceGroupId: `/spaces/${this.spaceGroupId}`,
```

notes-webapp/src/main/webapp/vue-app/notes/components/NotesOverview.vue
```diff
@@ -532,11 +532,11 @@ export default {
     },
     addNote() {
       if (!this.hasDraft) {
-        window.open(`${eXo.env.portal.context}/${eXo.env.portal.portalName}/notes-editor?spaceId=${eXo.env.portal.spaceId}&parentNoteId=${this.note.id}&appName=${this.appName}`, '_blank');
+        window.open(`${eXo.env.portal.context}/${eXo.env.portal.portalName}/notes-editor?spaceId=${eXo.env.portal.spaceId}&parentNoteId=${this.note.id}&spaceGroupId=${eXo.env.portal.spaceGroup}&appName=${this.appName}`, '_blank');
```
ahamdi: Could you add a nullity check for `eXo.env.portal.spaceGroup` to avoid having a null value ?

ahamdi (changes requested):
Could you update the description of the PR and rephrase it

ahamdi (approved):

PR #593, opened 2022-07-18
MEED-300: Add the global notification alert component for the note favorite button

SaraBoutej:
Prior to this change, we add a new component for notification alert when clicking on favorite button.
In this task we remove the current component and replace it by the global notification alert.


AzmiTouil (approved):
LGTM

PR #607, opened 2022-08-22
TASK-59276: Improve protection to protect REST services

hchaabene:
Prior this change, 2 rest API endpoints **exportNote()** and **cancelExportNote()** have @POST as HTTP method but doesn't write in DATABASE so they need to be changed to @GET

### DIFF
--- a/notes-service/src/main/java/org/exoplatform/wiki/service/rest/NotesRestService.java
+++ b/notes-service/src/main/java/org/exoplatform/wiki/service/rest/NotesRestService.java
@@ -806,10 +806,10 @@ public class NotesRestService implements ResourceContainer {
     }
   }
 
-  @POST
+  @GET
   @Path("/note/export/{exportId}/{notes}")
   @RolesAllowed("users")
-  @Operation(summary = "Export notes", method = "PUT", description = "This export selected notes and provide a zip file.")
+  @Operation(summary = "Export notes", method = "GET", description = "This export selected notes and provide a zip file.")
   @ApiResponses(value = { @ApiResponse(responseCode = "200", description = "Request fulfilled"),
       @ApiResponse(responseCode = "400", description = "Invalid query input"), @ApiResponse(responseCode = "403", description = "Unauthorized operation"),
       @ApiResponse(responseCode = "404", description = "Resource not found") })
@@ -879,7 +879,7 @@ public class NotesRestService implements ResourceContainer {
     }
   }
 
-  @PUT
+  @GET
   @Path("/note/export/cancel/{exportId}")
   @RolesAllowed("users")
   @Produces(MediaType.APPLICATION_JSON)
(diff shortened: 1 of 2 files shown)


notes-service/src/main/java/org/exoplatform/wiki/service/rest/NotesRestService.java
```diff
@@ -114,11 +114,11 @@ public NotesRestService(NoteService noteService,
     cc.setNoStore(true);
   }
 
-  @GET
+  @PUT
```
rdenarie: Why did you change this one ? By reading the function, it seems to be a simple read (getNote)

notes-service/src/main/java/org/exoplatform/wiki/service/rest/NotesRestService.java
```diff
@@ -307,11 +307,11 @@ public Response getLatestDraftOfPage(@Parameter(description = "Note id", require
     }
   }
 
-  @GET
+  @POST
```
rdenarie: This one seems also a READ action

rdenarie (changes requested):

azayati (changes requested):
@hchaabene I don't think that we should change annotations, this is not what is asked by the US

notes-service/src/main/java/org/exoplatform/wiki/service/rest/NotesRestService.java
```diff
@@ -118,7 +118,7 @@ public NotesRestService(NoteService noteService,
   @Path("/note/{noteBookType}/{noteBookOwner:.+}/{noteId}")
   @Produces(MediaType.APPLICATION_JSON)
   @RolesAllowed("users")
-  @Operation(summary = "Get note by notes params", method = "GET", description = "This get the not if the authenticated user has permissions to view the objects linked to this note.")
+  @Operation(summary = "Get note by notes params", method = "PUT", description = "This get the not if the authenticated user has permissions to view the objects linked to this note.")
```
azayati: revert this change

notes-service/src/main/java/org/exoplatform/wiki/service/rest/NotesRestService.java
```diff
@@ -311,7 +311,7 @@ public Response getLatestDraftOfPage(@Parameter(description = "Note id", require
   @Path("/versions/{noteId}")
   @Produces(MediaType.APPLICATION_JSON)
   @RolesAllowed("users")
-  @Operation(summary = "Get versions of note by id", method = "GET", description = "This get the versions of a note if the authenticated user has permissions to view the objects linked to this note.")
+  @Operation(summary = "Get versions of note by id", method = "POST", description = "This get the versions of a note if the authenticated user has permissions to view the objects linked to this note.")
```
azayati: revert this change

azayati (changes requested):
see review and change description according to the last review changes @hchaabene

azayati (approved):
LGTM

rdenarie (approved):

PR #610, opened 2022-09-12
 TASK-52149: Fix different titles style between edited notes and saved notes

akhanfir:
Prior to this change, when editing a note, heading 1,2 and 3 are not bold and when the note is saved they change to bold . To fix this we have changed the styles (font-size and font-weight) of text `h1`,`h2` and `h3`.


azayati (approved):
LGTM

PR #612, opened 2022-09-13
fix: Fix different titles style between edition and saved notes - TASK-52149 - Meeds-io/notes#610 

akhanfir:
Prior to this change, in edition heading 1,2 and 3 are not bold, and when the note is saved they change to bold . To fix this, change the style (font-size and font-weight) of text `h1`,`h2` and `h3`.


azayati (approved):
LGTM

PR #614, opened 2022-09-15
TASK-59276: Improve protection to protect REST services

hchaabene:
Prior this change, 2 rest API endpoints **exportNote()** and **cancelExportNote()** have https://github.com/post as HTTP method but doesn't write in DATABASE so they need to be changed to @get


azayati (approved):
LGTM

PR #616, opened 2022-09-21
feat: Re-use the common version history drawer - EXO-59522 - Meeds-io/MIPs#19

hakermi:
Re-use the common version history drawer already created in social commons


mkrout (approved):
LGTM

PR #617, opened 2022-09-23
fix: Fix export/import a note lead to disappearance of  images in it - TASK-58694 - Meeds-io/meeds-314

akhanfir:
Prior to this change, when we try to import an exported note with images in its content, the imported note images disappear.
After this change, we ensure to process the imported note content in order to generate the adequate images links.


notes-service/src/main/java/org/exoplatform/wiki/service/impl/NoteServiceImpl.java
```diff
@@ -1240,14 +1247,15 @@ public void importNote(Page note, Page parent, Wiki wiki, String conflict, Ident
             }
             note.setName(newTitle);
             note.setTitle(newTitle);
+            String processedContent = htmlUploadImageProcessor.processSpaceImages(note.getContent(), wiki.getOwner(), null);
             note.setContent(note.getContent());
```
azayati:
```suggestion
            note.setContent(processedContent);
```

azayati (changes requested):
see review

azayati (approved):
LGTM

PR #627, opened 2022-10-25
fix: Fix export/import a note lead to disappearance of images in it - EXO-58694 - Meeds-io/meeds-314

akhanfir:
Prior to this change, when we try to import an exported note with images in its content, the imported note images disappear. After this change, we ensure to process the imported note content in order to generate the adequate images links.


azayati (approved):
LGTM

PR #632, opened 2022-10-26
fix: Can't add an image for a new note - EXO-59684

Jihed525:
Prior to this change, when creating a new note and clicking on the upload image button from the insert plugin drawer of notes ckeditor, for brave browser the upload form isn't displayed and for FF browser the upload form is well displayed but the upload process does not finish correctly. After this commit, to fix the problem, w try to init the ckeditor when opening the notes custom plugins.

(cherry picked from commit 8a30165a3c00a96ced204643b71b1699cc7da0d1)


azayati (approved):
LGTM

PR #643, opened 2022-12-05
fix: History version drawer is empty - EXO-60449 - meeds#442

mkrout:
Prior to this fix the history version of a note that already had versions is totally empty. since for notes the id is not added to the object.


hakermi (approved):

rdenarie:
Can you add the meeds issue in the PR title ?
 https://github.com/Meeds-io/meeds/issues/442
Thx

PR #654, opened 2023-01-18
 fix: notes favIcon bug - EXO-61827

GouadriaHanen:
Prior to this change, Fav icon is colored in yellow even though these notes aren't bookmarked because the favorite-button component displays the first selected data.
after this change, the Added attribute key is used to force the changes when the data changes


ahamdi:
Could you update the title of the PR ?

ahamdi (approved):

PR #659, opened 2023-02-07
fix: fix displaying empty version drawer - EXO-62185

GouadriaHanen:
before this modification, when opening the versions drawer from the dropdown menu, the drawer was empty because no versionPageSize was defined
after this modification, the version page size is defined (in this function openNoteVersionsHistoryDrawer) and the drawer is opened and the versions are displayed successfully.

### DIFF
--- a/notes-webapp/src/main/webapp/vue-app/notes/components/NotesOverview.vue
+++ b/notes-webapp/src/main/webapp/vue-app/notes/components/NotesOverview.vue
@@ -239,7 +239,7 @@
       :default-path="defaultPath"
       @open-treeview="$refs.notesBreadcrumb.open(note, 'movePage')"
       @export-pdf="createPDF(note)"
-      @open-history="openNoteVersionsHistoryDrawer(noteVersions, isManager)"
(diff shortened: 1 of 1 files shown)


notes-webapp/src/main/webapp/vue-app/notes/components/NotesOverview.vue
```diff
@@ -239,7 +239,7 @@
       :default-path="defaultPath"
       @open-treeview="$refs.notesBreadcrumb.open(note, 'movePage')"
       @export-pdf="createPDF(note)"
-      @open-history="$refs.noteVersionsHistoryDrawer.open(noteVersions,isManager)"
+      @open-history="openNoteVersionsHistoryDrawer(noteVersions, isManager)"
```
ahamdi: function openNoteVersionsHistoryDrawer does not take parameters, could you clean the invocation ?

ahamdi (changes requested):

ahamdi (approved):

PR #732, opened 2023-09-19
fix: Adjust blockquote wysiwyg styles - EXO-65725 - Meeds-io/MIPs#71

hakermi:
Adjust blockquote wysiwyg styles


mkrout (approved):

rdenarie (changes requested):
This modification only applies on notes. Is it possible to make it for all ckeditor in the platform ? 
I'm afraid that this modification creates something not consistent

hakermi:
> This modification only applies on notes. Is it possible to make it for all ckeditor in the platform ? I'm afraid that this modification creates something not consistent

@rdenarie you can discuss this with @Julien-Dubois-eXo .
I know, all what i'm doing is only and specific for notes, that's what i have asked to do,  you can refer and discuss this with the POs of the ft

PR #843, opened 2023-11-13
Merge ckeditor improvements - Meeds-io/MIPs#71

### LINKED ISSUE #71: Improvement in the editor of note app
## Rationale

- Increase the CK Editor capacity in Notes as many customers requests more options
- Optimise the use of iFramely to correctly include it in Notes. 
- Additionally, iFramely spotted us and blocked Builder and Tribe.
- Avoid the redaction at the same time and the loss of data. 

### 1.1 Editor options

**To add**: 

- Emojis (same as in the stream composer)
- Tags (same as in the stream composer)
- Decrease indent
- Increase indent
- Code snippet
- Anchor
- Accessibility checker

**To rework:**

- **Table**: Remove the current component and use the standard CK Editor component. To remove from the Insert option drawer. Lock the table width to the screen width (we can't have a scroll inside the note)
- **Link**: Remove the current component and use the standard CK Editor component.
  * Info link standard
  * Target: self; blank; undefined
- **Insert media**: Remove the current "Insert video" component and use the standard CK Editor component "Media embedded" based on iFramely

- **Right click**: 
  Correct to have the standard menu. 

Image

Image

-  **WYSIWYG**: rework the layout (size, style, interline, etc) to be sure that the layout is the same in the editor and the viewer

<html>
<body>

Element | Layout
-- | --
Page | Editor: - _Width_: 1321.620px **> 1232px** - Margin left/right: 36px  _View_ - Width: 1232px - Margin left/right: 24px **> 36px**
Paragraph | Style: Normal - Font size: 18,667px **> 16px** - Line-height: 1.4 (= 1,4xFont size =26.1334px) - Interline: 0px - Space before a line: 0px - Space after a line: 10px
Title 1 | Font size: 34px **> 28px** Font-weight: 400 **> 500** Line-height: 1.2 = 1,2xFont size =px) Interline: 0px Space before: 10px **> 45px** Space after: 10px
Title 2 | Font size: 28px **> 23px** Font-weight: 400 **> 500** Line-height: 1.2 = 1,2xFont size =px) Interline: 0px Space before: 10px **> 35px** Space after: 10px
Title 3 | Font size: 21,84px **> 18px** Font-weight: 400 **> 500** Line-height: 1.2 = 1,2xFont size =px) Interline: 0px Space before: 10px **> 25px** Space after: 10px
Bullet point list | Font size: 18,667px **>16px** Line-height: 20px **>1.4** (specific for Notes) Interline: 0px Space before a line: 0px Space after a line: 0px **Space after the list:10px** Before Total indent: 25px **> 40px**  - bullet point center in the indent? - indent before bullet: (40/2)-bullet size - indent before text: (40/2)-bullet size After padding: 25px **> 40px** **Exit list: Normal** **Change the bullet point when increasing indentation**
Numbered list | Font size: 18,667px **>16px** Line-height:20px **>1.4** (specific for Notes) Interline: 0px Space before a line: 0px Space after a line: 0px **Space after the list:10px** Before total indent: 25px **> 40px**  - number center in the indent? - indent before number: (40/2)-bullet size - indent before text: (40/2)-bullet size After padding: 25px **> 40px ** **Exit list: Normal**
Quote | Font-size: 17,5px **> 16px** Interline: 0px font-weight: 300 **> 400** Space before a line: 0px Space after a line: 0px Space after the quote: 10px
Table | Space before the table: 0px **Space after the table: 10px** **Text inside: Normal**
Image | Marges should be those of the editor.
Embedded media | Space before the table: 0px **Space after the table: 10px**
Note link | Style: Normal
URL link | Style: Normal
Table of content | Font size: 14px **> 16px** **Open in a new tab option with a right click on 3rd mouse button**

</body>
</html>

Image

## 2. Technical Requirements

Mainely requirements are about customizing the notes style and adding new CKeditor plugins:

- [ ]  The required styles changes will be only applied on Notes side so to be able to redifine some styles for notes and to insure extending & reusing the CKEditor Stylesheet & Javascript instead of redefining them.

### CK editor side (edit mode)

**In the Commons project:**

The"commons-extension-webapp" will be used to:
- The content.css containing the ck editor styles will be renamed to content.less ( to allow using Less extensibility).
- In the content.less file all values that need to be customized should be repalced with variables (the list of css styles to be defined)
- And a new "variables.less" file to be added contanning defult values for variables used in the "content.less" file
- An "custom-variables.less" to be added, this file will be used by higher layers modules to overide the variables 

**In the notes project:**

- Add "custom-variables.less" file containing all variable to be redifined.
- Use io.meeds.social.core.richeditor.RichEditorConfigurationPlugin to define a custom Javascript configuration file for notes CKEditor instances
- Add the overriden CSS in Javascript : config.contentsCss.push('/notes/ckeditorCustom/contents.css'); instead of config.contentsCss.push('/commons-extension/ckeditorCustom/contents.css');
(for more details check the POC provided here: https://github.com/Meeds-io/MIPs/issues/71#issuecomment-1746267640)

### Notes app side (view mode):

**In the Commons project:**

- In the "commons-extension-webapp" add a "extendedRichContent.less" file were all styles used to ensure WYSIWYG should be used with variables defined in "variables.less" for CK editor.
- The content.less file shoulld be imported under the "extended-rich-content" css class in "extendedRichContent.less" 
```
    .extended-rich-content {
      @import "...../......./content.less";
      .......
    }
```
- Any needed css styles for extended view will be added under "extended-rich-content" css class in the "extendedRichContent.less".
- As done for the Ck editor side, a css file will be created from extendedRichContent.less using the "variables.less"  and the "custom-variables.less"

**In the notes side:**

- import the css file generated from the richEditorContent.less (using the predefined custom-variables.less) in the notes.less file.
- add the "extended-rich-content" class after "rich-editor-content" to override the default styles.

- [ ] New addons will be added to commons:
- Link
- copyFormatting
- colorDialog
- codensippets
- table

### Expected Volume & Performance
No impact on volume or performance

### Security 
No security impact

hakermi:
Merge ckeditor improvements


notes-service/src/main/java/org/exoplatform/wiki/jpa/search/WikiElasticSearchServiceConnector.java
```diff
@@ -70,6 +70,15 @@ public class WikiElasticSearchServiceConnector extends ElasticSearchServiceConne
 
   private String                     searchQueryFilePath;
 
+  public static final String           SEARCH_QUERY_TERM            = ",\"must\":{" +
+          "  \"query_string\":{" +
+          "    \"fields\": [\"name\",\"title\",\"content\",\"comment\",\"attachment.content\"]," +
+          "    \"default_operator\": \"AND\"," +
+          "    \"query\": \"@term@\"" +
+          "  }" +
+          "}";
```
boubaker: You can use `triple double quote sequence` to simplify reading and maintenance:

```suggestion
  public static final String           SEARCH_QUERY_TERM            = """
         ,"must":{
            "query_string":{
              "fields": ["name","title","content","comment","attachment.content"],
              "default_operator": "AND",
              "query": "@term@"
            }
          }
  """;
```

notes-service/src/main/java/org/exoplatform/wiki/jpa/search/WikiElasticSearchServiceConnector.java
```diff
@@ -111,33 +120,67 @@ protected String getSourceFields() {
...
     String jsonResponse = getClient().sendRequest(esQuery, getIndex());
     return buildWikiResult(jsonResponse);
   }
 
+  private String buildTagsQueryStatement(List<String> values) {
+    if (values == null || values.isEmpty()) {
+      return "";
+    }
+    List<String> tagsQueryParts = values.stream()
+            .map(value -> "{\"term\": {\n" +
```
boubaker: Same, you can use `triple double quote sequence` to simplify reading and maintenance

notes-service/src/main/java/org/exoplatform/wiki/listener/AnalyticsAddTagListener.java
```diff
@@ -0,0 +1,49 @@
+package org.exoplatform.wiki.listener;
```
boubaker: Missing license Header
(To apply on all classes please to not repeat the comment multiple times)

boubaker: For new classes in Meeds, please use package name `io.meeds` instead of `org.exoplatform`
(To apply on all classes please to not repeat the comment multiple times)

```suggestion
package io.meeds.wiki.listener;
```

notes-service/src/main/java/org/exoplatform/wiki/service/impl/NoteServiceImpl.java
```diff
@@ -132,6 +132,7 @@ public class NoteServiceImpl implements NoteService {
 
   private final HTMLUploadImageProcessor                  htmlUploadImageProcessor;
 
+
```
boubaker: useless line break
```suggestion
```

notes-service/src/main/java/org/exoplatform/wiki/service/impl/NoteServiceImpl.java
```diff
@@ -242,13 +242,13 @@ public Page createNote(Wiki noteBook, String parentNoteName, Page note, Identity
...
       postAddPage(noteBook.getType(), noteBook.getOwner(), note.getName(), createdPage);
-
+      Utils.broadcast(listenerService, "note.posted", userIdentity.getUserId(), createdPage);
       return createdPage;
     } else {
       throw new EntityNotFoundException("Parent note not foond");
     }
   }
-
+
```
boubaker: useless space

```suggestion
```

notes-service/src/main/java/org/exoplatform/wiki/service/impl/NoteServiceImpl.java
```diff
@@ -275,6 +275,7 @@ public Page updateNote(Page note, PageUpdateType type, Identity userIdentity) th
       }
     }
     Page updatedPage = updateNote(note, type);
+    Utils.broadcast(listenerService, "note.updated", userIdentity.getUserId(), updatedPage);
```
boubaker: Shouldn't we trigger/centralize this event in method `public Page updateNote(Page note, PageUpdateType type)`, WDYT ?

notes-service/src/main/java/org/exoplatform/wiki/utils/Utils.java
```diff
@@ -684,4 +692,11 @@ public static String html2text(String html) {
     return doc.text();
   }
 
+  public static <S, D> void broadcast(ListenerService listenerService, String eventName, S source, D data) {
+    try {
+      listenerService.broadcast(eventName, source, data);
+    } catch (Exception e) {
+      LOG.error("Error while broadcasting event: {}", eventName, e);
```
boubaker: Use warn log level when the operation will not break the rest of computation. This will be used just for logging and troubleshooting.

```suggestion
      LOG.warn("Error while broadcasting event: {}. Wheras, the operation will continue to proceed", eventName, e);
```

notes-webapp/src/main/webapp/WEB-INF/conf/wiki/metadata-configuration.xml
```diff
@@ -0,0 +1,42 @@
+<!--
+    Copyright (C) 2023 eXo Platform SAS.
```
boubaker: Wrong license Header

```suggestion
```

notes-webapp/src/main/webapp/vue-app/notes/components/NotesOverview.vue
```diff
@@ -199,8 +199,8 @@
             :items="noteAllChildren"
             item-key="noteId">
             <template #label="{ item }">
-              <v-list-item-title @click="openNoteChild(item)" class="body-2 clickable primary--text">
-                <span>{{ item.name }}</span>
+              <v-list-item-title class="body-2 content-link clickable primary--text">
+                <a :href="getNoteLink(item.noteId)"><span>{{ item.name }}</span></a>
```
boubaker: For a common UX with other parts of the product, can you use please `<v-list-item :href` instead of `<v-list-item-title....<a :href`

```suggestion
              <v-list-item :href="getNoteLink(item.noteId)" class="body-2 primary--text">
                {{ item.name }}
              </v-list-item>
```

Besides, to not use a method here and use a computed instead, can you add a dedicated component for Tree leaf.

notes-webapp/src/main/webapp/vue-app/notes/components/NotesOverview.vue
```diff
@@ -809,13 +813,17 @@ export default {
       const contentChildren = docElement.getElementsByTagName('body')[0].children;
       const links = docElement.getElementsByTagName('a');
       const tables = docElement.getElementsByTagName('table');
+      docElement.querySelectorAll('pre > code').forEach((e) => codeHighlighter.highlightBlock(e));
```
boubaker: Using a 149KB of new framework + its corresponding CSS seems not relevant. Please try to delete the usage of this framework if it's not a major functional requirement, at least in "**View Mode**".

```suggestion
```

notes-service/src/main/java/org/exoplatform/wiki/service/impl/NoteServiceImpl.java
hakermi: I used the updatePage function that is called in the rest layer in all update actions while the above method is used only in the service layer.
And the important reason is that i need the `userIdentity` which is not available in this method.

notes-webapp/src/main/webapp/vue-app/notes/components/NotesOverview.vue
hakermi: > Using a 149KB of new framework + its corresponding CSS seems not relevant. Please try to delete the usage of this framework if it's not a major functional requirement, at least in "**View Mode**".

This library comes with ckeditor and used in the codesnippet plugin, so i created an amd library of it + a portal skin of the choose style of the code to ensure the WYSIWYG for codesnippet because there is no other way to do that in view mode. ckidetor.getData() doesn't return the exactly highlighted html code.
https://github.com/Meeds-io/commons/pull/551/commits/90092cc73ff14e478395b4f844cd14ed5c152ddf#diff-60986d51a51cfa41dd1f469bd45a68212c4512d265ab7dfb946a3127b6671862R5

notes-service/src/main/java/org/exoplatform/wiki/service/impl/NoteServiceImpl.java
boubaker: Yep, this is not the good pattern to use for event triggering .... When the update id made by a batch update (auto import when server starts up by example), the listener should be triggered as well if if no user had auctioned the update.

notes-webapp/src/main/webapp/vue-app/notes/components/NotesOverview.vue
boubaker: ok for Style but no need to make the framework work in **View Mode**
The plugin is intended to generate an HTML snippet that has to be saved as is. And when doing so, no need to retrigger the plugin in View mode to have a good display of the injected HTML.

hakermi: Ok but without the userId the listener is useless
 https://github.com/Meeds-io/notes/pull/843/files#diff-3198ed87ff299273c36b60c054d7cfdef2b2eada4a3f90bd54d51fe498237cb7R61-R62

notes-service/src/main/java/org/exoplatform/wiki/service/impl/NoteServiceImpl.java
hakermi: Ok but without the userId the listener is useless
https://github.com/Meeds-io/notes/pull/843/files#diff-3198ed87ff299273c36b60c054d7cfdef2b2eada4a3f90bd54d51fe498237cb7R61-R62

notes-webapp/src/main/webapp/vue-app/notes/components/NotesOverview.vue
hakermi: OK but even if we keep the styles, in the view mode we will not have the same display of the code as in edit mode, as all highlighted parts are transformed into span tags with css classes, but these span tags are removed when call editor.getData()

notes-service/src/main/java/org/exoplatform/wiki/jpa/search/WikiElasticSearchServiceConnector.java
hakermi: done

notes-service/src/main/java/org/exoplatform/wiki/jpa/search/WikiElasticSearchServiceConnector.java
hakermi: done

notes-service/src/main/java/org/exoplatform/wiki/listener/AnalyticsAddTagListener.java
hakermi: done

hakermi: done

notes-service/src/main/java/org/exoplatform/wiki/service/impl/NoteServiceImpl.java
hakermi: done

notes-service/src/main/java/org/exoplatform/wiki/service/impl/NoteServiceImpl.java
hakermi: done

notes-webapp/src/main/webapp/WEB-INF/conf/wiki/metadata-configuration.xml
hakermi: fixed

notes-webapp/src/main/webapp/vue-app/notes/components/NotesOverview.vue
hakermi: done and to check

notes-service/src/main/java/org/exoplatform/wiki/utils/Utils.java
hakermi: done

notes-webapp/src/main/webapp/vue-app/notes/components/NotesOverview.vue
mkrout: And if we save the generated HTML, we will not be able to edit it again.

notes-service/src/main/java/org/exoplatform/wiki/service/impl/NoteServiceImpl.java
boubaker: Is it really?

```java
    Page note = event.getData();
    String username = event.getSource();
    long creatorId = getPosterId(username);
    long audienceId = getStreamOwnerId(note.getWikiOwner(), username);

    Set<TagName> tagNames = tagService.detectTagNames(note.getContent());
    tagService.saveTags(new TagObject(NOTES_METADATA_OBJECT_TYPE, note.getId(), note.getParentPageId()),
                        tagNames,
                        audienceId,
                        creatorId);
    indexingService.reindex(WikiPageIndexingServiceConnector.TYPE, note.getId());
```

The listener should:
* update Tags
* reindex the page

**independently** whether the action is made **by a user** or **by a batch process**, those updates has to be made.

Shouldn't we delete the `indexingService.reindex` since the note page is automatically reindexed when a metadata item is modified (triggered using listener `org.exoplatform.wiki.service.listener.MetadataItemModified`) ?

notes-webapp/src/main/webapp/vue-app/notes/components/NotesOverview.vue
boubaker: > And if we save the generated HTML, we will not be able to edit it again.

We have to not use `HTMLSanitizer` anymore when writing the information. (we don't have to loose the original information entered by the end user)
But we use `v-sanitized-html` to sanitize the html on rendering in order to preserve the original information.

notes-service/src/main/java/org/exoplatform/wiki/service/impl/NoteServiceImpl.java
hakermi: I didn't understand,
```
 tagService.saveTags(new TagObject(NOTES_METADATA_OBJECT_TYPE, note.getId(), note.getParentPageId()),
                        tagNames,
                        audienceId,
                        creatorId
```); 

In the createTags isn't the userIdentityId mandatory ? 
how i can get it without passing it as parameter in the event?

Comment (boubaker) [reply notes-service/src/main/java/org/exoplatform/wiki/service/impl/NoteServiceImpl.java]: you can use the last updater or the creator, it's not really important information that will be used after saving the associated Tags to a note, the most important information is the `audienceId`

File: notes-webapp/src/main/webapp/vue-app/notes/components/NotesOverview.vue (hunk 10 above)
Comment (hakermi) [reply notes-webapp/src/main/webapp/vue-app/notes/components/NotesOverview.vue]: The problem is not in the `HtmlSanitizer` it's in the ckeditor itself, in the `editor.getData()`

Comment (boubaker) [reply notes-webapp/src/main/webapp/vue-app/notes/components/NotesOverview.vue]: > And if we save the generated HTML, we will not be able to edit it again.

>The problem is not in the HtmlSanitizer it's in the ckeditor itself, in the editor.getData()

This is may be a cross cutting concern about CKEditor contents. We may need to preserve the HTML as displayed in CKEditor (`CKEDITOR.instances.**INSTANCE_NAME**.document.$.body.innerHTML`) to preserve the WYSIWYG effect, not only for notes. At the same time, we will need to preserve the response for `editor.getData()` to be able to edit the content with CKEditor plugins.

This would be a good alternative to [PHP backend processor](https://ckeditor.com/docs/ckeditor4/latest/features/codesnippetgeshi.html#back-end-configuration) of a version of this plugin.

We will talk more about this in the dedicated meeting.

File: notes-service/src/main/java/org/exoplatform/wiki/service/impl/NoteServiceImpl.java (hunk 6 above)
Comment (hakermi) [reply notes-service/src/main/java/org/exoplatform/wiki/service/impl/NoteServiceImpl.java]: There is no creator or updater property in the note page and the `page.getAuthor()` is always null because there is no author property in `pageEntity`

File: notes-webapp/src/main/webapp/vue-app/notes/components/NotesOverview.vue (hunk 10 above)
Comment (hakermi) [reply notes-webapp/src/main/webapp/vue-app/notes/components/NotesOverview.vue]: I have already used something similar to preserve the embedded iframe of the media plugin
https://github.com/Meeds-io/notes/pull/843/files#diff-43927e45889ed109fe9f553ea38b02993f822a8fb1827fa9d26c660fa5cf3c79R813-R834
should i try to do the same for highlighter too ?

Comment (boubaker) [reply notes-webapp/src/main/webapp/vue-app/notes/components/NotesOverview.vue]: The idea is to minimize the **CPU & Memory consumption**, in browser side, when displaying Notes in Page. It should be as much static as possible to optimize the necessary ressources to display a page with multiple **SNVs** based on Notes.
Having to load a **heavy JS library** in addition to a computing to transform the HTML each time we display the page wouldn't be performant.

File: notes-service/src/main/java/org/exoplatform/wiki/service/impl/NoteServiceImpl.java (hunk 6 above)
Comment (boubaker) [reply notes-service/src/main/java/org/exoplatform/wiki/service/impl/NoteServiceImpl.java]: > There is no creator or updater property in the note page and the `page.getAuthor()` is always null because there is no author property in `pageEntity`

I let you find a solution, for my part the flow and layers are important to preserve coherence and information consistency.

File: notes-webapp/src/main/webapp/vue-app/notes/components/NotesOverview.vue (hunk 10 above)
Comment (hakermi) [reply notes-webapp/src/main/webapp/vue-app/notes/components/NotesOverview.vue]: @boubaker PR updated
https://github.com/Meeds-io/notes/pull/843/commits/86836dc8b9e90bb08c614b8282ac7f50314ef4f2

File: notes-service/src/main/java/io/meeds/notes/listener/NotesMetadataListener.java [hunk 11]
```diff
@@ -0,0 +1,90 @@
...
+    this.tagService = tagService;
+    this.userACL = userACL;
+  }
+
+  @Override
+  public void onEvent(Event<String, Page> event) {
+    Page note = event.getData();
+    String username = event.getSource();
+    if (username == null) {
+      username = userACL.getSuperUser();
```
boubaker, line 65: Can you use `note.getAuthor()` first if it's not null + ensure to populate the attribute when updating/creating it.
When author is null as well (like when importing a note from sources in server startup), in this special case, we will have to use `username = userACL.getSuperUser();`

notes-service/src/main/java/io/meeds/notes/listener/NotesMetadataListener.java
```diff
@@ -0,0 +1,90 @@
...
+    }
+    long creatorId = getPosterId(username);
+    long audienceId = getStreamOwnerId(note.getWikiOwner(), username);
+
+    Set<TagName> tagNames = tagService.detectTagNames(note.getContent());
+    tagService.saveTags(new TagObject(NOTES_METADATA_OBJECT_TYPE, note.getId(), note.getParentPageId()),
+                        tagNames,
+                        audienceId,
+                        creatorId);
+    indexingService.reindex(WikiPageIndexingServiceConnector.TYPE, note.getId());
```
boubaker: Please split this listener into two different listeners, each one will consider one Task to do at the same time. Using this combination of Tasks inside a single Listener will lead to not index the content when the Tags throws an error. This is a **best practice** the fact to split into multiple listeners and avoid combining them.

notes-service/src/main/java/org/exoplatform/wiki/service/impl/NoteServiceImpl.java
```diff
@@ -242,7 +242,7 @@ public Page createNote(Wiki noteBook, String parentNoteName, Page note, Identity
 
       // call listeners
       postAddPage(noteBook.getType(), noteBook.getOwner(), note.getName(), createdPage);
-
+      Utils.broadcast(listenerService, "note.posted", note.getLastUpdater(), createdPage);
```
boubaker: We should broadcast listeners as well in `public Page createNote(Wiki noteBook, Page parentPage, Page note)` to ensure information consistency & coherence even in batch method.

notes-service/src/main/java/org/exoplatform/wiki/service/impl/NoteServiceImpl.java
```diff
@@ -297,15 +297,8 @@ public Page updateNote(Page note, PageUpdateType type) throws WikiException {
...
-    if (PageUpdateType.EDIT_PAGE_CONTENT.equals(type) || PageUpdateType.EDIT_PAGE_CONTENT_AND_TITLE.equals(type)) {
-      try {
-        listenerService.broadcast("exo.wiki.edit", wikiService, note);
-      } catch (Exception e) {
-        log.error("Error while broadcasting wiki edition event", e);
-      }
-    }
-
     Page updatedPage = getNoteById(note.getId());
+    Utils.broadcast(listenerService, "note.updated", note.getLastUpdater(), updatedPage);
```
boubaker: We should broadcast listeners as well in `public Page updateNote(Page note)` to ensure information consistency & coherence even in batch method.

notes-webapp/src/main/webapp/vue-app/notes/components/NoteContentTableItem.vue
```diff
@@ -0,0 +1,23 @@
...
+  props: {
+    note: {
+      type: Object,
+      default: null
+    }
+  },
+  computed: {
+    noteLink() {
+      const baseUrl = window.location.href;
+      return this.note && `${baseUrl.substring(0, baseUrl.lastIndexOf('/') + 1)}${this.note.noteId}`;
```
boubaker: We should use the `url` attribute returned in REST response. If the REST doesn't return it, we should fix the REST endpoint.

```suggestion
      return this.note?.url;
```

notes-webapp/src/main/webapp/vue-app/notes/components/NotesOverview.vue
```diff
@@ -809,13 +807,16 @@ export default {
       const contentChildren = docElement.getElementsByTagName('body')[0].children;
       const links = docElement.getElementsByTagName('a');
       const tables = docElement.getElementsByTagName('table');
+      docElement.querySelectorAll('oembed').forEach((oembed) => {
+        oembed.innerHTML = oembed.dataset.iframe;
+        delete oembed.dataset.iframe;
+      });
```
boubaker: Sorry, I didn't get why we still need to change the saved HTML to make the display works. Can't we get a complete HTML from REST endpoint and display it as is ?
We will have a whole **CMS set of features** on top of this work. We should take the time to improve the performances of contents rendering now, to not have to make heavy and a substantial work for data migration just after few months. If we don't optimize the maximum of things now, we will probably never prioritize the performance improvements later when we will see that the page loading takes a lot of time when it's full of **SNVs**.

notes-service/src/main/java/io/meeds/notes/listener/NotesMetadataListener.java
hakermi: note.getAuthor() is always null i said previously, because there is not author property in the pageEntity

boubaker: yes, as agreed @mkrout @rdenarie in meeting, we should populate it with the last author name.

notes-service/src/main/java/io/meeds/notes/listener/NotesMetadataListener.java
hakermi: you want me to remove the `indexingService.reindex(...` right ? OK i will

notes-service/src/main/java/io/meeds/notes/listener/NotesMetadataListener.java
hakermi: I already used a lastUpdater attribute and send it in the broadcast why i should use the author ?
`Utils.broadcast(listenerService, "note.posted", **note.getLastUpdater()**, createdPage);`

notes-webapp/src/main/webapp/vue-app/notes/components/NotesOverview.vue
hakermi: Same as codeSnippets plugin, the cke media plugin needs a processing as well, because when we insert a link using the embed media plugin, in edit mode  all is OK, but when call `editor.getData()` we have only "`<oembed>usedLink</oembed>`"

notes-service/src/main/java/io/meeds/notes/listener/NotesMetadataListener.java
hakermi: I have removed the reindex because it happens in the `MetaDataModified` listener even if it doesn't make sense that when the `detectTags` fails(**task1**) we should `reindex`(**task2**).

notes-webapp/src/main/webapp/vue-app/notes/components/NoteContentTableItem.vue
hakermi: I have fixed the null url, but the whole idea about replacing with list-item has killed the ui of the table content.
beside a black color that is set in vuetify-all with !important flag in the list item there is also a default grey hover effect

notes-webapp/src/main/webapp/vue-app/notes/components/NotesOverview.vue
hakermi: I have had to do something about this : https://github.com/Meeds-io/notes/pull/843/commits/a71cb13e8f8313ba82d6e6d17e232759bfdf8ba7#diff-43927e45889ed109fe9f553ea38b02993f822a8fb1827fa9d26c660fa5cf3c79R811-R852

`restoreEmbedded()` and `restoreunHighlited`() are needed to make the content compatible again with ckeditor in edit mode and recognized by the corresponding plugins

notes-service/src/main/java/io/meeds/notes/listener/NotesMetadataListener.java
boubaker: even if it's already covered at this time by the caller method, this method should be auto sufficient and as much coherent as possible by embedding inside it the full business logic, which is:
1. If event.getData() isn't null, then use it
2. Else if page.getAuthor isn't null, then use it
3. Else, when the page is new and was added by a batch method (like import pages process), use a default author which is the super admin

This way we are sure to have secured the business logic even when a future maintenance/feature is made on the caller method where the author isn't passed as event data object.

notes-service/src/main/java/org/exoplatform/wiki/model/Page.java
```diff
@@ -98,6 +98,8 @@ public class Page {
   private Page                            parent;
 
   private Map<String, List<MetadataItem>> metadatas;
+  
+  private String                          lastUpdater;
```
boubaker: As agreed in a meeting with @mkrout and @rdenarie we will reuse the `author` instead to have the last modifier, can we use it instead ?

notes-webapp/src/main/webapp/vue-app/notes-editor/components/NotesEditorDashboard.vue
```diff
@@ -853,9 +808,67 @@ export default {
...
+          const height = iframes[index]?.parentNode?.offsetHeight;
+          const aspectRatio = width / height;
+          const minHeight = parseInt(this.oembedMinWidth) / aspectRatio;
+          const style = `
+          min-height: ${minHeight}px;
+          min-width: ${this.oembedMinWidth}px;
+          width: 100%;
+          margin-bottom: 10px;
+          aspect-ratio: ${aspectRatio};
+        `;
```
boubaker: Formatting for code clearness

```suggestion
            min-height: ${minHeight}px;
            min-width: ${this.oembedMinWidth}px;
            width: 100%;
            margin-bottom: 10px;
            aspect-ratio: ${aspectRatio};
          `;
```

notes-webapp/src/main/webapp/vue-app/notes/components/NoteContentTableItem.vue
boubaker: > I have fixed the null url, but the whole idea about replacing with list-item has killed the ui of the table content.

Yes, that's the idea of using href on `v-list-item`, see it in multiple UIs, like Top bar menu:

image

This aspect is to validate with PMs, I let you discuss about this internally and decide which style to use.

But for the URL retrieving from REST endpoint instead of building it here, knowing that we already have this in REST while getting a single note, would be better for code coherence. In fact, getting a single `NotePage` of a list of `NotePage`s should retrieve the same set of information, except when we use a different `expand` parameter.

notes-webapp/src/main/webapp/skin/less/notes/notes.less
```diff
@@ -7,7 +7,8 @@
 @import "Notifications/Style.less";
 @import "Sprites/Style.less";
 @import "UIWikiRichTextEditor/Stylesheet.less";
-
+@import "../../../ckeditor/skins/common/extendedRichContent.less";
+@import (less) "../../../ckeditor/plugins/codesnippet/lib/highlight/styles/monokai_sublime.css";
```
boubaker, line 11: Not sure that this is the best place to define the reusable Less files, since we have platform-ui for it. I will change it anyway when starting to work on SNVs V2. Nothing to do more from your part, I 'll take it from here, but just to share the best practice for future implementations.

notes-webapp/src/main/webapp/skin/less/notes/notes.less
```diff
@@ -60,18 +61,20 @@
     z-index: 1031;
     background: #eeeeee;
 
+    .notes-content-wrapper {
+      min-height: calc(~"100vh - 210px");
```
boubaker: there is new way introduced to manage Mobile ViewPort bug ([see it here](https://medium.com/@alekswebnet/fix-mobile-100vh-bug-in-one-line-of-css-dynamic-viewport-units-in-action-102231e2ed56))

```suggestion
      min-height: ~"calc(var(--100vh, 100vh) - 210px)";
```

notes-webapp/src/main/webapp/skin/less/notes/notes.less
```diff
@@ -60,18 +61,20 @@
...
+    }
+
     .notes-content {
-      height: calc(~"100vh - 120px");
+      height: calc(~"100vh - 100px");
```
boubaker: same, even if we shouldn't use the viewport multiple times in the same page, but define it in the parent container and then use flexbox to adjust using `flex-grow-1`, `flex-grow-0`, `flex-column`...

```suggestion
      height: ~"calc(var(--100vh, 100vh) - 100px)";
```

notes-webapp/src/main/webapp/skin/less/notes/notes.less
```diff
@@ -60,18 +61,20 @@
...
       overflow: auto;
       z-index: 5;
       background-color: #eeeeee;
       margin-top: 102px;
 
       .notes-content-form {
-        background: white;
-        margin: 20px 14% 0 14% !important;
+        margin: 20px auto !important;
```
boubaker: Can you avoid defining margins in CSS and move this into HTML using `mx-.my.ma.mt-...`

notes-webapp/src/main/webapp/skin/less/notes/notes.less
```diff
@@ -341,6 +291,12 @@
       }
     }
   }
+
+  .content-link {
+    a {
+      font-size: 16px !important;
```
boubaker: We already have CSS helpers to define font sizes of elements, can you use it please instead.
Imagine we want to change the fonts sizes globally, we will have to change in one single place, plus this strategy avoids to have a huge CSS files to parse in browser side (CPU & Memory) and to download (Network performances)

notes-webapp/src/main/webapp/skin/less/notes/notes.less
hakermi: here i can't have exactly 20px with veutify helpers

notes-webapp/src/main/webapp/skin/less/notes/notes.less
hakermi: which helper that allow to set exactly 16px as fontsize ?

notes-service/src/main/java/io/meeds/notes/listener/NotesMetadataListener.java
boubaker: > it doesn't make sense that when the detectTags fails(task1) we should reindex(task2).

Shouldn't we have the content of the updated note searchable when the Tags aren't correct ? Two different Use Cases that shouldn't block each other.

notes-webapp/src/main/webapp/vue-app/notes/components/NotesOverview.vue
boubaker: ok, no need to do more, we have to re-architecture the whole stuff to enhance this, by adding by example an additional field in DB for `Content To Edit` & keep the current one for `Content To View`. This would be the good integration to make with CKEditor advanced plugins that was added by current MIP... we will need to rework this but may be in the next MIP. Let's just keep it a simple implementation for now... Thanks.

notes-service/src/main/java/io/meeds/notes/listener/NotesMetadataListener.java
```diff
@@ -0,0 +1,91 @@
...
+    this.userACL = userACL;
+  }
+
+  @Override
+  public void onEvent(Event<String, Page> event) {
+    Page note = event.getData();
+    String username = event.getSource();
+    if (username == null && note.getAuthor() != null) {
+      username = note.getAuthor();
+    } else if (username == null) {
```
boubaker: To simplify code clearness, thus to simplify code maintenance

```suggestion
    if (username == null) {
      username = note.getAuthor();
    }
    if (username == null) {
```

notes-webapp/src/main/webapp/skin/less/notes/notes.less
boubaker: `subtitle-1` defines in rem defines the style of subtitles used in product.

notes-service/src/main/java/org/exoplatform/wiki/service/impl/NoteServiceImpl.java
```diff
@@ -251,13 +250,16 @@ public Page createNote(Wiki noteBook, String parentNoteName, Page note, Identity
 
   @Override
   public Page createNote(Wiki noteBook, Page parentPage, Page note) throws WikiException {
-
-    return dataStorage.createPage(noteBook, parentPage, note);
+    Page createdPage = dataStorage.createPage(noteBook, parentPage, note);
+    Utils.broadcast(listenerService, "note.posted", note.getAuthor(), createdPage);
```
boubaker, line 254: We should trigger `postAddPage` as well

notes-service/src/main/java/org/exoplatform/wiki/service/impl/NoteServiceImpl.java
```diff
@@ -251,13 +250,16 @@ public Page createNote(Wiki noteBook, String parentNoteName, Page note, Identity
...
+    return createdPage;
   }
 
   @Override
   public Page updateNote(Page note) throws WikiException {
-    return dataStorage.updatePage(note);
+    Page updatedPage = dataStorage.updatePage(note);
+    Utils.broadcast(listenerService, "note.updated", note.getAuthor(), updatedPage);
```
boubaker: We should trigger `postUpdatePage` as well

notes-service/src/main/java/org/exoplatform/wiki/service/impl/NoteServiceImpl.java
```diff
@@ -284,6 +287,7 @@ public Page updateNote(Page note, PageUpdateType type, Identity userIdentity) th
     updatedPage.setAppName(note.getAppName());
     Map<String, List<MetadataItem>> metadata = retrieveMetadataItems(note.getId(), userIdentity.getUserId());
     updatedPage.setMetadatas(metadata);
+    postUpdatePage(updatedPage.getWikiType(), updatedPage.getWikiOwner(), updatedPage.getName(), updatedPage, type);
```
boubaker, line 297: Already triggered in common method `updateNote(note)`, so we should avoid to trigger it twice

```suggestion
```

notes-service/src/main/java/org/exoplatform/wiki/service/rest/NotesRestService.java
```diff
@@ -384,6 +384,7 @@ public Response createNote(@Parameter(description = "note object to be created",
       note.setSyntax(syntaxId);
       note.setName(TitleResolver.getId(note.getTitle(), false));
       note.setUrl("");
+      note.setAuthor(currentUser);
```
boubaker: Should be centralized in Service layer, since the identity ACL is passed to Service (remember, in REST, no business logic should be done, which unfortunately not the case of a lot of stuff in this REST endpoint, but at least we need to avoid making it worse)

```suggestion
```

notes-service/src/main/java/org/exoplatform/wiki/service/impl/NoteServiceImpl.java
hakermi: In the other method there is no type param which is needed in the GamificationListener 
https://github.com/Meeds-io/notes/blob/develop/notes-social-integration/src/main/java/org/exoplatform/wiki/integration/gamification/GamificationWikiListener.java#L99

notes-webapp/src/main/webapp/vue-app/notes/components/NoteContentTableItem.vue
```diff
@@ -0,0 +1,24 @@
+<template>
```
boubaker, line 21: Missing license header

notes-service/src/main/java/org/exoplatform/wiki/service/impl/NoteServiceImpl.java
boubaker: In this case we centralize the real implementation of update in method `updateNote(Page note, PageUpdateType type)` (with **storage** and **listeners** call) and change the implementation of `updateNote(Page note)` this way :

```java
  public Page updateNote(Page note) throws WikiException {
    return updateNote(note, null);
  }
```

boubaker: in fact, it would even be better to centralize all in the **full method** `Page updateNote(Page note, PageUpdateType type, Identity userIdentity)` and change the others this way:

```java
  public Page updateNote(Page note) throws WikiException {
    return updateNote(note, null);
  }

  public Page updateNote(Page note, PageUpdateType type) throws WikiException {
    return updateNote(note, type, null);
  }
```

boubaker:
Ok for me, thanks. Can you please squash the new commits added by PR review and add a significant title for each chunk of modifications.

use commit message of type :
`fix: ..... - Meeds-io/MIPs#71`

boubaker (approved):

PR #866, opened 2023-11-23
fix: Cleanup Notes API for better maintainability - MEED-2980 - Meeds-io/meeds#1308

boubaker:


boubaker:
Will be reworked after 1.5.0-GA release

PR #873, opened 2023-11-24
Merge Notes multilingual feature

mkrout:


boubaker (approved):
Nothing blocker, some huge Vue file, license header, formatting and missing fine Error handling in REST endpoint especially for ACL check. But let's merge this, no need to keep  working on it knowing that it will be improved later when working on SNV.

PR #907, opened 2023-12-22
Fix: Notes - Navigation menu in a note with a text is not displayed in view mode - EXO-67951 - Meeds-io/meeds#1484

mkrout:
Before this commit when user add text to a note and insert navigation widget, the navigation does not appear in the view mode. This commit fix the problem


azayati (approved):
LGTM

PR #911, opened 2023-12-26
fix: Prevent image dragging on the single note view application - EXO-68249

sofyenne:
This change is going to prevent image dragging on the single note view application


notes-webapp/src/main/webapp/vue-app/note-page-view/components/NoteRichEditor.vue
```diff
@@ -119,6 +119,7 @@ export default {
           copyFormatting_allowedContexts: true,
           isImagePasteBlocked: true,
           hideUploadImageLink: true,
+          isImageDragBlocked: true,
```
ahamdi, line 122: can we use imageDragDisabled ?

ahamdi (approved):

PR #915, opened 2023-12-27
fix: Fix scrolling issue in the single note view application edit mode - EXO-68518_68519

IlhemEssaadi:
Prior to this change, scrolling when editing in the single note view application  was not functioning as expected. This commit adds configuration option `autoGrow_maxHeight` to improve scrolling functionality.

### DIFF
--- a/notes-webapp/src/main/webapp/vue-app/note-page-view/components/NoteRichEditor.vue
+++ b/notes-webapp/src/main/webapp/vue-app/note-page-view/components/NoteRichEditor.vue
@@ -117,6 +117,8 @@ export default {
           enterMode: CKEDITOR.ENTER_P,
           shiftEnterMode: CKEDITOR.ENTER_BR,
           copyFormatting_allowedContexts: true,
+          isImagePasteBlocked: true,
+          hideUploadImageLink: true,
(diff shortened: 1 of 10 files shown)


notes-webapp/src/main/webapp/vue-app/note-page-view/components/NoteRichEditor.vue
```diff
@@ -123,6 +123,8 @@ export default {
             offset: 40,
             unit: 'px'
           },
+          autoGrow_onStartup: true,
+          autoGrow_maxHeight: 800,
```
boubaker: Should be able to change this in the dedicated file of configuration:

https://github.com/Meeds-io/notes/blob/83e6286aec1054f9f614d7b3d96ad542b0ffe847/notes-webapp/src/main/webapp/WEB-INF/conf/wiki/ckeditor/inline-config.js#L55-L56

boubaker: ok for me, thanks. I resolve this comment and let you continue the review with the team.

azayati (approved):
LGTM

PR #935, opened 2024-02-09
fix: Prevent the auto import of the knowledge base space if it already exist - EXO-69419

sofyenne:
Prior to this change after renaming one of the knowledge base space and restart the exo server , the knowledge base space would be recreated , this issue arose due to the space's pretty name being issued to check it's existence , which was changed by renaming the space , This change is going to check the space's existence using the space group id resolving this issue .

### DIFF
--- a/notes-service/src/main/java/org/exoplatform/wiki/service/NotesAutoImportService.java
+++ b/notes-service/src/main/java/org/exoplatform/wiki/service/NotesAutoImportService.java
@@ -193,8 +193,8 @@ public class NotesAutoImportService implements Startable {
                            String spaceDescription,
                            String zipPath,
                            Identity superUserIdentity) {
-
-    Space space = spaceService.getSpaceByPrettyName(spaceName);
+    String groupId  = "/spaces/" + spaceName;
+    Space space = spaceService.getSpaceByGroupId(groupId);
     if (space == null) {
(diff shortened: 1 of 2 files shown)


rdenarie (changes requested):
Can we also change the property name which have a typo in the xml file ? 

https://github.com/Meeds-io/notes/blob/9420218988f49395160f4c0c15b7e64c726f38ed/notes-service/src/main/resources/conf/portal/configuration.xml#L93

it should be exo.notes.knowledge.base.fr.space.name

sofyenne:
> Can we also change the property name which have a typo in the xml file ?
> 
> https://github.com/Meeds-io/notes/blob/9420218988f49395160f4c0c15b7e64c726f38ed/notes-service/src/main/resources/conf/portal/configuration.xml#L93
> 
> it should be exo.notes.knowledge.base.fr.space.name

done!

sofyenne:
> Can we also change the property name which have a typo in the xml file ?
> 
> https://github.com/Meeds-io/notes/blob/9420218988f49395160f4c0c15b7e64c726f38ed/notes-service/src/main/resources/conf/portal/configuration.xml#L93
> 
> it should be exo.notes.knowledge.base.fr.space.name

done!

rdenarie (approved):

PR #946, opened 2024-03-26
feat: Allow to create a root note page other that the homepage - EXO-70331 - Meeds-io/MIPs#119

sofyenne:
This change is going to  allow to create a root note page other that the homepage .


azayati (approved):
LGTM

PR #956, opened 2024-04-22
feat: Remove jspdf and html2canvas declared amd libraries -EXO - Meeds-io/MIPs#112

hakermi:
prior to this change, notes declare its dedicated jspdf and html2canvas amd libraries,
This PR removes this declaration to rely on the declared libraries is social module


azayati (approved):
LGTM

PR #966, opened 2024-04-29
fix: Delete orphan draft pages when the space is deleted - EXO-70385 - Meeds-io/MIPs#119

hakermi:
Prior to this change, when deleting a note space, the related space note data is not well deleted when the space have orphan drafts (for non existing pages) which raises an exception related to the constraint foreign key violation when attempting to delete note pages with children orphan drafts.
This PR makes sure to delete all note drafts before proceeding to delete the related note pages.


azayati (approved):
LGTM

PR #973, opened 2024-05-13
fix: Fix add translation in sub note when no translation connector active - EXO-71327 - Meeds-io/meeds#1987

hakermi:
prior to this change, issue when adding translation in sub note when no translation connector active because of the default returned note when no translation exists for requested language. 
This PR fixes the issue by adding a check to make sure that we have requested for a specific lang version note and that wasn't exists.


azayati (approved):
LGTM

PR #985, opened 2024-05-29
fix: Navigation menu not displayed if text added in the body of a note - EXO-71919 - Meeds-io/meeds#2067

hakermi:
Navigation menu not displayed if text added in the body of a note


azayati (approved):
LGTM

PR #1001, opened 2024-06-10
fix: No Image properties option in editor context menu - EXO-72349 - Meeds-io/meeds#2142

hakermi:
Restore the Image properties option in the editor contextmenu


azayati (approved):
LGTM

PR #1005, opened 2024-06-13
fix: Adjust icons alignment in notes drop down menu - EXO-72270 - Meeds-io/MIPs#128

hakermi:
Adjust icons alignment in notes drop down menu


azayati (approved):
LGTM

PR #1019, opened 2024-07-01
feat: Implement save note and review validate editor form - EXO-71943_EXO-71949

hakermi:
Save note and validate editor form review


azayati (approved):
LGTM

PR #1020, opened 2024-07-01
fix: fix editor header icons and related ui/ux - EXO-71931 - Meeds-io/MIPs#128

hakermi:
fix editor header icons and related ui/ux


azayati (approved):
LGTM

PR #1022, opened 2024-07-02
fix: Allow add/edit notes in mobile - EXO-72857 - Meeds-io/MIPs#128

hakermi:
Allow add/edit notes in mobile

### DIFF
--- a/notes-webapp/src/main/webapp/vue-app/notes/components/NotesOverview.vue
+++ b/notes-webapp/src/main/webapp/vue-app/notes/components/NotesOverview.vue
@@ -21,7 +21,7 @@
               class="notes-header-icons text-right">
               <div
                 class="d-inline-flex">
-                <v-tooltip bottom v-if="!isMobile && !hasDraft && isManager">
+                <v-tooltip bottom v-if="!hasDraft && isManager">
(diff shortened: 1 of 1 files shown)


notes-webapp/src/main/webapp/vue-app/notes/components/NotesOverview.vue
```diff
@@ -21,7 +21,7 @@
               class="notes-header-icons text-right">
               <div
                 class="d-inline-flex">
-                <v-tooltip bottom v-if="!isMobile && !hasDraft && isManager">
+                <v-tooltip bottom v-if="!hasDraft && isManager">
```
azayati, line 24: can you check why before the revamping, it was working with this condition !isMobile ?

azayati (changes requested):

notes-webapp/src/main/webapp/vue-app/notes/components/NotesOverview.vue
hakermi: The old mobile breakpoint wasn't working, that's why

azayati (approved):
LGTM

PR #1023, opened 2024-07-02
fix: Use save label in the editor action button for both update/save actions - EXO-71949 - Meeds-io/MIPs#128

hakermi:
Change the the Update button label from update to save


azayati (approved):
LGTM

PR #1024, opened 2024-07-03
fix: Prevent redirect to view mode when create a note - EXO-71943 - Meeds-io/MIPs#128

hakermi:
Prevent redirect to view mode when create a note


azayati (approved):
LGTM

PR #1034, opened 2024-07-10
fix: Cannot update a note during creation -  EXO-73062 - Meeds-io/MIPs#128

hakermi:
Cannot update a note during creation


azayati (approved):
LGTM

PR #1050, opened 2024-07-23
feat: Implement mentions in note user notification - EXO-73071 - Meeds-io/MIPs#128

sofyenne:


notes-service/src/main/java/org/exoplatform/wiki/service/impl/NoteServiceImpl.java
```diff
@@ -283,6 +291,10 @@ public Page createNote(Wiki noteBook, Page parentPage, Page note) throws WikiExc
 
     Utils.broadcast(listenerService, "note.posted", note.getAuthor(), createdPage);
     postAddPage(noteBook.getType(), noteBook.getOwner(), note.getName(), createdPage);
+    Matcher mentionMatcher = MentionInNoteNotificationPlugin.MENTION_PATTERN.matcher(createdPage.getContent());
```
ahamdi: Could you avoid depending of MentionInNoteNotificationPlugin here.
Could you use some static class for constants or create one if needed.

sofyenne: done!

ahamdi (approved):

PR #1060, opened 2024-08-06
feat: Review Note display - EXO-71927 - Meeds-io/MIPs#128

hakermi:
Review Note display

### DIFF
--- a/notes-webapp/src/main/webapp/skin/less/notes/notes.less
+++ b/notes-webapp/src/main/webapp/skin/less/notes/notes.less
@@ -309,11 +309,24 @@
     min-height: calc(~"100vh - 170px");
 
     .notes-application-header {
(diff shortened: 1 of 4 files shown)


notes-webapp/src/main/webapp/skin/less/notes/notes.less
```diff
@@ -309,11 +309,22 @@
     min-height: calc(~"100vh - 170px");
 
     .notes-application-header {
+      .note-actions-menu {
+        right: 35px;
+        top: 50px !important;
+        left: auto !important;
```
ahamdi: Could you consider the RTL display here ?

hakermi: Done

ahamdi (approved):

PR #1069, opened 2024-08-23
fix: Adapt Thumbnail Service API change - Meeds-io/MIPs#128

azayati:
After merging MIP-134, the class ImageThumbnailService.java location has been changed from social module to gatein-portal module with a package modification, in addition to that the used method getOrCreateThumbnail signature has been changed also.
After this change, the notes code will be adapted according to theses modifications.


rdenarie:
Can you Change the commit message ? In fact, this commit, even if fixing a modification which occurs in mip-134 is in reality related to mip 128. So we should have a commit message relative to this mip like this : 
""fix: Adapt Thumbnail Service API change - Meeds-io/MIPs#128"
WDYT ?

rdenarie (changes requested):
change commit message please

hbenali (dismissed):

rdenarie (approved):

PR #1074, opened 2024-08-27
fix: Adjust ui/ux issues in note view mode - EXO-73673 - Meeds-io/MIPs#128

hakermi:
Adjust ui/ux issues in note view mode


azayati (approved):
LGTM

PR #1077, opened 2024-08-28
feat: Include properties in note export/import - EXO-73694 - Meeds-io/MIPs#129

hakermi:
Include properties in note export/import


azayati (approved):
LGTM

PR #1078, opened 2024-08-29
fix: Editor header disappears once clicked on translate button right after note creation - EXO-73506 - Meeds-io/MIPs#128

hakermi:
Editor header disappears once clicked on translate button right after note creation


azayati (approved):
LGTM

PR #1079, opened 2024-08-29
fix: Fix accessibility issues in open/close metadata drawer buttons - EXO-73675 - Meeds-io/MIPs#128

hakermi:
fix accessibility issues in open/close metadat drawer buttons

### DIFF
--- a/notes-webapp/src/main/webapp/vue-app/notes-rich-editor/components/NoteEditorTopBar.vue
+++ b/notes-webapp/src/main/webapp/vue-app/notes-rich-editor/components/NoteEditorTopBar.vue
@@ -61,8 +61,9 @@
             min-width="36"
             height="36"
             class="pa-0 my-auto "
-            icon
+            :aria-label="$t('notes.metadata.open.drawer')"
             :disabled="!editorReady"
+            icon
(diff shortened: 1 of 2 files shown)


notes-webapp/src/main/webapp/vue-app/notes-rich-editor/components/NoteEditorTopBar.vue
```diff
@@ -107,6 +108,7 @@
               width="36"
               min-width="36"
               height="36"
+              :aria-label="$t('notes.metadata.close.drawer')"
```
azayati: we need to make the close button of properties drawer accessible not the close button of the editor on mobile

azayati (changes requested):

notes-webapp/src/main/webapp/vue-app/notes-rich-editor/components/NoteEditorTopBar.vue
hakermi: yes, you're right, but we don't have such option in the drawer

azayati (approved):
LGTM

PR #1091, opened 2024-09-04
fix: Featured image link is not updated in note view when navigate using treeview - EXO-73961 - Meeds-io/MIPs#128

hakermi:
Featured image link is not updated in note view when navigate using treeview


azayati (approved):
LGTM

PR #1094, opened 2024-09-06
feat: Remove metadata properties when delete note/draft - EXO-73990 - Meeds-io/MIPs#128

hakermi:
Remove metadata properties when delete note/draft

### DIFF
--- a/notes-service/src/main/java/org/exoplatform/wiki/jpa/JPADataStorage.java
+++ b/notes-service/src/main/java/org/exoplatform/wiki/jpa/JPADataStorage.java
@@ -20,20 +20,6 @@
 
 package org.exoplatform.wiki.jpa;
 
-import static org.exoplatform.wiki.jpa.EntityConverter.convertAttachmentEntityToAttachment;
-import static org.exoplatform.wiki.jpa.EntityConverter.convertAttachmentToDraftPageAttachmentEntity;
-import static org.exoplatform.wiki.jpa.EntityConverter.convertAttachmentToPageAttachmentEntity;
-import static org.exoplatform.wiki.jpa.EntityConverter.convertDraftPageEntityToDraftPage;
-import static org.exoplatform.wiki.jpa.EntityConverter.convertDraftPageToDraftPageEntity;
-import static org.exoplatform.wiki.jpa.EntityConverter.convertPageEntityToPage;
-import static org.exoplatform.wiki.jpa.EntityConverter.convertPageToPageEntity;
-import static org.exoplatform.wiki.jpa.EntityConverter.convertPageVersionEntityToPageHistory;
-import static org.exoplatform.wiki.jpa.EntityConverter.convertPageVersionEntityToPageVersion;
-import static org.exoplatform.wiki.jpa.EntityConverter.convertPermissionEntitiesToPermissionEntries;
-import static org.exoplatform.wiki.jpa.EntityConverter.convertPermissionEntriesToPermissionEntities;
-import static org.exoplatform.wiki.jpa.EntityConverter.convertWikiEntityToWiki;
-import static org.exoplatform.wiki.jpa.EntityConverter.convertWikiToWikiEntity;
-
 import java.io.UnsupportedEncodingException;
 import java.net.URLEncoder;
 import java.util.ArrayList;
@@ -105,7 +91,9 @@ import org.exoplatform.wiki.utils.NoteConstants;
 import org.exoplatform.wiki.utils.Utils;
 import org.exoplatform.wiki.utils.VersionNameComparatorDesc;
 
-/**
+import static org.exoplatform.wiki.jpa.EntityConverter.*;
+
+ /**
  * Created by The eXo Platform SAS Author : eXoPlatform exo@exoplatform.com
  * 9/8/15
  */
@@ -382,15 +370,14 @@ public class JPADataStorage implements DataStorage {
 
   @Override
   public void deleteDraftOfPage(Page page) throws WikiException {
-    List<DraftPageEntity> draftPages = draftPageDAO.findDraftPagesByTargetPage(Long.parseLong(page.getId()));
-    for (DraftPageEntity draftPage: draftPages) {
-      if(draftPage != null){
-        deleteAttachmentsOfDraftPage(draftPage);
-      }
-    }
-    draftPageDAO.deleteDraftPagesByTargetPage(Long.valueOf(page.getId()));
+    draftPageDAO.deleteDraftPagesByTargetPage(Long.parseLong(page.getId()));
   }
 
+  @Override
+  public void deleteAttachmentsOfDraftPage(DraftPage draftPage) {
+    deleteAttachmentsOfDraftPage(convertDraftPageToDraftPageEntity(draftPage, pageDAO));
+  }
+  
   @Override
   public void deleteDraftOfPage(Page page, String lang) throws WikiException {
     List<DraftPageEntity> draftPages = draftPageDAO.findDraftPagesByTargetPage(Long.parseLong(page.getId()));
@@ -679,16 +666,10 @@ public class JPADataStorage implements DataStorage {
     }
     return null;
   }
-  
+
   @Override
   public List<DraftPage> getDraftsOfPage(Long pageId) {
-    List<DraftPage> draftPages = new ArrayList<>();
-    List<DraftPageEntity> draftPagesOfUser = draftPageDAO.findDraftPagesByTargetPage(pageId);
-    for (DraftPageEntity draft : draftPagesOfUser) {
-      draftPages.add(convertDraftPageEntityToDraftPage(draft));
-    }
-
-    return draftPages;
+    return convertDraftPageEntitiesToDraftPages(draftPageDAO.findDraftPagesByTargetPage(pageId));
   }
   
   @Override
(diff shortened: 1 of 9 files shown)


notes-service/src/main/java/org/exoplatform/wiki/service/impl/NoteServiceImpl.java
```diff
@@ -907,6 +901,17 @@ public void removeDraftOfNote(Page page, String username) throws WikiException {
    */
   @Override
   public void removeDraftOfNote(Page page) throws WikiException {
+    List<DraftPage> draftPages = dataStorage.findDraftPagesByTargetPage(page);
```
azayati:
```suggestion
    List<DraftPage> draftPages = dataStorage.getDraftsOfPage(Long.parseLong(page.getId()));
```
IMO you can use the existing method dataStorage.getDraftsOfPage instead of adding a new one

notes-service/src/main/java/org/exoplatform/wiki/jpa/JPADataStorage.java
```diff
@@ -382,15 +370,19 @@ private void deletePageEntity(PageEntity pageEntity) {
...
-      if(draftPage != null){
-        deleteAttachmentsOfDraftPage(draftPage);
-      }
-    }
-    draftPageDAO.deleteDraftPagesByTargetPage(Long.valueOf(page.getId()));
+    draftPageDAO.deleteDraftPagesByTargetPage(Long.parseLong(page.getId()));
   }
 
+  @Override
+  public List<DraftPage> findDraftPagesByTargetPage(Page page) {
```
azayati: IMO you can use the existing method dataStorage.getDraftsOfPage instead of adding a new one

notes-service/src/main/java/org/exoplatform/wiki/service/DataStorage.java
```diff
@@ -92,6 +92,14 @@ public interface DataStorage {
 
   public void deleteDraftOfPage(Page page) throws WikiException;
 
+  /**
+   * Finds list of associated drafts by target page id
+   * 
+   * @param page target page
+   * @return {@link List} of {@link DraftPage}
+   */
+  List<DraftPage> findDraftPagesByTargetPage(Page page);
```
azayati: IMO you can use the existing method dataStorage.getDraftsOfPage instead of adding a new one

notes-service/src/main/java/org/exoplatform/wiki/service/impl/NoteServiceImpl.java
```diff
@@ -453,7 +447,7 @@ public boolean deleteNote(String noteType, String noteOwner, String noteName, Id
         log.error("Can't delete note '" + noteName + "'. This note does not exist.");
         return false;
       }
-    } catch (WikiException e) {
+    } catch (Exception e) {
```
azayati:
```suggestion
    } catch (WikiException e) {
```
I think here that the only catched exception is WikiException

notes-service/src/main/java/org/exoplatform/wiki/jpa/JPADataStorage.java
```diff
@@ -382,15 +370,19 @@ private void deletePageEntity(PageEntity pageEntity) {
 
   @Override
   public void deleteDraftOfPage(Page page) throws WikiException {
-    List<DraftPageEntity> draftPages = draftPageDAO.findDraftPagesByTargetPage(Long.parseLong(page.getId()));
-    for (DraftPageEntity draftPage: draftPages) {
-      if(draftPage != null){
-        deleteAttachmentsOfDraftPage(draftPage);
```
azayati, line 388: IMO as it works well currently no need to move deleteAttachmentsOfDraftPage to the service layer

notes-service/src/main/java/org/exoplatform/wiki/service/impl/NoteServiceImpl.java
```diff
@@ -907,6 +901,17 @@ public void removeDraftOfNote(Page page, String username) throws WikiException {
    */
   @Override
   public void removeDraftOfNote(Page page) throws WikiException {
+    List<DraftPage> draftPages = dataStorage.findDraftPagesByTargetPage(page);
+    for (DraftPage draftPage : draftPages) {
+      if (draftPage != null) {
+        dataStorage.deleteAttachmentsOfDraftPage(draftPage);
```
azayati, line 907: see above comment

notes-service/src/main/java/org/exoplatform/wiki/service/impl/NoteServiceImpl.java
```diff
@@ -943,15 +948,11 @@ public void removeDraft(String draftName) throws WikiException {
    */
   @Override
   public void removeDraftById(String draftId) throws Exception {
-    Page draftNote = dataStorage.getDraftPageById(draftId);
+    DraftPage draftNote = dataStorage.getDraftPageById(draftId);
+    deleteNoteMetadataProperties(draftNote, draftNote.getLang(), NOTE_METADATA_DRAFT_PAGE_OBJECT_TYPE);
```
azayati, line 952: I prefer that the properties deletion come after the draft deletion itself, so when there is a problem with the draft deletion the properties will not be deleted

notes-service/src/main/java/org/exoplatform/wiki/service/impl/NoteServiceImpl.java
```diff
@@ -907,6 +901,17 @@ public void removeDraftOfNote(Page page, String username) throws WikiException {
    */
   @Override
   public void removeDraftOfNote(Page page) throws WikiException {
+    List<DraftPage> draftPages = dataStorage.findDraftPagesByTargetPage(page);
+    for (DraftPage draftPage : draftPages) {
+      if (draftPage != null) {
+        dataStorage.deleteAttachmentsOfDraftPage(draftPage);
+        try {
+          deleteNoteMetadataProperties(draftPage, draftPage.getLang(), NOTE_METADATA_DRAFT_PAGE_OBJECT_TYPE);
```
azayati, line 909: I prefer that the properties deletion come after the draft deletion itself, so when there is a problem with the draft deletion the properties will not be deleted

notes-service/src/main/java/org/exoplatform/wiki/service/impl/NoteServiceImpl.java
```diff
@@ -1466,22 +1469,46 @@ public DraftPage getLatestDraftPageByUserAndTargetPageAndLang(Long targetPageId,
    * {@inheritDoc}
    */
   @Override
-  public void deleteVersionsByNoteIdAndLang(Long noteId, String lang) throws WikiException {
+  public void deleteVersionsByNoteIdAndLang(Long noteId, String lang) throws Exception {
+    Page note = getNoteById(String.valueOf(noteId));
+    if (note != null) {
+      deleteNoteMetadataProperties(note, lang, NOTE_METADATA_PAGE_OBJECT_TYPE);
```
azayati, line 1475: I prefer that the properties deletion come after the version deletion itself, so when there is a problem with the version deletion the properties will not be deleted

notes-service/src/main/java/org/exoplatform/wiki/service/impl/NoteServiceImpl.java
```diff
@@ -2085,11 +2113,6 @@ private Map<String, String> copyNotePageProperties(Page oldNote,
       MetadataItem newNoteMetadataItem = getNoteMetadataItem(note, lang, newObjectType);
       if (oldNoteMetadataItem != null && oldNoteMetadataItem.getProperties() != null) {
         properties = oldNoteMetadataItem.getProperties();
-        if (properties != null && oldLang == null && lang != null) {
-          properties.remove(FEATURED_IMAGE_ID);
-          properties.remove(FEATURED_IMAGE_ALT_TEXT);
-          properties.remove(FEATURED_IMAGE_UPDATED_DATE);
-        }
         if (newNoteMetadataItem != null) {
           newNoteMetadataItem.setProperties(properties);
```
azayati: no more need to properties initialization IMO, you can use directly newNoteMetadataItem.setProperties(oldNoteMetadataItem.getProperties())

azayati (changes requested):

notes-service/src/main/java/org/exoplatform/wiki/jpa/JPADataStorage.java
hakermi: IMO it should be moved to avoid two calls + two loops on drafts

notes-service/src/main/java/org/exoplatform/wiki/service/impl/NoteServiceImpl.java
hakermi: see above reply

notes-service/src/main/java/org/exoplatform/wiki/service/impl/NoteServiceImpl.java
hakermi: Nope, we need to delete properties before, because when we delete them we need to fetch the draft/note an compare illustrations use, so we don't want to fetch a deleted draft

notes-service/src/main/java/org/exoplatform/wiki/service/impl/NoteServiceImpl.java
hakermi: https://github.com/Meeds-io/notes/pull/1094#discussion_r1753509503

notes-service/src/main/java/org/exoplatform/wiki/service/impl/NoteServiceImpl.java
hakermi: https://github.com/Meeds-io/notes/pull/1094#discussion_r1753509503

notes-service/src/main/java/org/exoplatform/wiki/jpa/JPADataStorage.java
azayati: in that case we should be sure that deleteAttachmentsOfDraftPage is called for all removeDraft methods

azayati (approved):
LGTM

azayati (approved):
LGTM

PR #1095, opened 2024-09-09
FIX: Notes tags - click on tags on home page displays no results - EXO-73907

mkrout:
Prior to this fix, search on for home notes doesn't display any note, this is due to that home page don't have poster. This commit we will use the last modifier user in the search result.


Jihed525 (approved):

Jihed525:
@mkrout it should be content on develop, no ?

azayati:
should this be related to a meeds issue @mkrout ?

PR #1099, opened 2024-09-10
feat: Include note properties in import/export notes - EXO-73694 - Meeds-io/MIPs#129

hakermi:
Include note properties in import/export notes

### DIFF
--- a/notes-service/src/main/java/org/exoplatform/wiki/service/ExportThread.java
+++ b/notes-service/src/main/java/org/exoplatform/wiki/service/ExportThread.java
@@ -20,15 +20,7 @@
 
 package org.exoplatform.wiki.service;
 
-import java.io.BufferedWriter;
-import java.io.ByteArrayInputStream;
-import java.io.File;
-import java.io.FileInputStream;
-import java.io.FileNotFoundException;
-import java.io.FileOutputStream;
-import java.io.FileWriter;
-import java.io.IOException;
-import java.io.InputStream;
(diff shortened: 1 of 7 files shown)


notes-service/src/main/java/org/exoplatform/wiki/service/ExportThread.java
```diff
@@ -144,6 +139,37 @@ public static File zipFiles(String zipFileName,
     return zip;
   }
 
+  private static void addFilesToZip(ZipOutputStream zipOut,
```
azayati: The private methods should be at the end after public ones, reorganize the whole class methods

notes-service/src/main/java/org/exoplatform/wiki/service/impl/NoteServiceImpl.java
```diff
@@ -1696,9 +1697,14 @@ protected void invalidateAttachmentCache(Page note) {
   
   /******* Private methods *******/
   
-  private void importNote(Page note, Page parent, Wiki wiki, String conflict, Identity userIdentity) throws WikiException,
-                                                                                                    IllegalAccessException, Exception {
-
+  private void importNote(Page note,
```
azayati, line 1915: reorganize the whole class methods (public, protected, private)

azayati (changes requested):

azayati (approved):
LGTM

PR #1101, opened 2024-09-10
feat: Disable property drawer in mobile/webPageNote - EXO-72335 - meeds-io/MIPs#128

hakermi:
Disbale property drawer in mobile/webPageNote


azayati (approved):
LGTM

PR #1107, opened 2024-09-16
fix: Properties of newly created note are not updated - EXO-74190 - Meed-io/MIPs#128

hakermi:
Properties of newly created note are not updated


azayati (approved):
LGTM

PR #1108, opened 2024-09-16
fix: Ui/ux issues in note create/switch translation - EXO-73222, EXO-74157 - Meeds-io/MIPs#129

hakermi:
Ui/ux issues in note create/switch translation:

- featured image not reset when automatic translation is not available while creating a new translation
- "draft saving..." message remains displayed when adding a new translation while the automatic translation is not available
- switch between translations enables the save button without existing of draft


azayati (approved):
LGTM

PR #1111, opened 2024-09-17
feat: Add a limit on note summary property value - EXO-74215 - Meeds-io/MIPs#129

hakermi:
Add a limit on note summary property value


azayati (approved):
LGTM

PR #1116, opened 2024-09-18
fix: Fix issue when update only summary of newly created note - EXO-74245 - Meeds-io/MIPs#128

hakermi:
fix Issue when update only summary of newly created note


azayati (approved):
LGTM

PR #1118, opened 2024-09-19
feat: Displaying the note title limit warning message only when trying to add more than the limit - EXO-74261 - Meeds-io/MIPs#129

sofyenne:


azayati (approved):
LGTM

PR #1134, opened 2024-09-26
fix: Fix issues in the listed note drafts treeview - EXO-74426 - Meeds-io/MIPs#129

hakermi:
Issues in listed note drafts treeview:
1 - fix the drafts tree build
2 - fix the unneeded displayed drafts of non targeted home pages

### DIFF
--- a/notes-service/src/main/java/org/exoplatform/wiki/jpa/JPADataStorage.java
+++ b/notes-service/src/main/java/org/exoplatform/wiki/jpa/JPADataStorage.java
@@ -1595,7 +1595,24 @@ public class JPADataStorage implements DataStorage {
    * {@inheritDoc}
    */
   @Override
-  public List<DraftPage> getDraftsOfWiki(String wikiOwner, String wikiType) {
-    return convertDraftPageEntitiesToDraftPages(draftPageDAO.findDraftsOfWiki(wikiOwner, wikiType));
+  public List<DraftPage> getDraftsOfWiki(String wikiOwner, String wikiType, String wikiHome) {
+    // The Note API allows multiple home pages to be created within the wiki.
+    // To avoid retrieving drafts from all home pages,
+    // we need to specifically fetch all pages under the target home page and search
+    // for drafts.
+    // Although this approach may seem performance-intensive, it is the only
+    // reliable solution.
+    PageEntity pageEntity = pageDAO.getPageOfWikiByName(wikiType, wikiOwner, wikiHome);
+    List<DraftPageEntity> draftPageEntities = new ArrayList<>();
+    getDraftsOfPage(pageEntity, draftPageEntities);
+    return convertDraftPageEntitiesToDraftPages(draftPageEntities);
   }
-}
+
+  private void getDraftsOfPage(PageEntity pageEntity, List<DraftPageEntity> drafts) {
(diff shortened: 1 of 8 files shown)


notes-service/src/main/java/org/exoplatform/wiki/jpa/JPADataStorage.java
```diff
@@ -1595,7 +1595,24 @@ public PageVersion getPageVersionById(long versionId) {
    * {@inheritDoc}
    */
   @Override
-  public List<DraftPage> getDraftsOfWiki(String wikiOwner, String wikiType) {
-    return convertDraftPageEntitiesToDraftPages(draftPageDAO.findDraftsOfWiki(wikiOwner, wikiType));
+  public List<DraftPage> getDraftsOfWiki(String wikiOwner, String wikiType, String wikiHome) {
+    // The Note API allows multiple home pages to be created within the wiki.
```
azayati, line 1599: normally only one home note is allowed for a wiki

notes-service/src/main/java/org/exoplatform/wiki/jpa/JPADataStorage.java
```diff
@@ -1595,7 +1595,24 @@ public PageVersion getPageVersionById(long versionId) {
...
+    // Although this approach may seem performance-intensive, it is the only
+    // reliable solution.
+    PageEntity pageEntity = pageDAO.getPageOfWikiByName(wikiType, wikiOwner, wikiHome);
+    List<DraftPageEntity> draftPageEntities = new ArrayList<>();
+    getDraftsOfPage(pageEntity, draftPageEntities);
+    return convertDraftPageEntitiesToDraftPages(draftPageEntities);
+  }
+
+  private void getDraftsOfPage(PageEntity pageEntity, List<DraftPageEntity> drafts) {
+    drafts.addAll(draftPageDAO.findDraftPagesByTargetPage(pageEntity.getId()));
```
azayati, line 1612: do we retrieve all drafts of a note page related to all languages ?

notes-webapp/src/main/webapp/vue-app/notes/components/NoteTreeviewDrawer.vue
```diff
@@ -709,6 +709,12 @@ export default {
       }
       return itemsArray;
     },
+    naturalSort(items) {
```
azayati, line 712: can we get the drafts sorted directly from back without sorting them in front ?

azayati (changes requested):

notes-service/src/main/java/org/exoplatform/wiki/jpa/JPADataStorage.java
hakermi: what is the Articles page created in content addon ?

notes-webapp/src/main/webapp/vue-app/notes/components/NoteTreeviewDrawer.vue
hakermi: It is used only in case of drafts, i don't see a reason to not do that in the front since i cant do it in request

azayati (approved):
LGTM

PR #1142, opened 2024-10-01
fix: Toc is not displayed in default empty note content - EXO-74504 - Meeds-io/MIPs#129

hakermi:
Toc is not displayed in default empty note content


azayati (approved):
LGTM

PR #1143, opened 2024-10-01
fix: Draft of new page is not correctly deleted after saving its page - EXO-74503 - Meeds-io/MIPs#129

hakermi:
Draft of new page is not correctly deleted after saving its page


azayati (approved):
LGTM

PR #1152, opened 2024-10-07
feat: Implement notes publication drawer - EXO-72738_EXO-72741 - Meeds-io/MIPs#161

hakermi:
Implement note publication drawer


azayati (changes requested):
Some remarks:

- The FF should be in the content not notes in order to disable/enable the display of the new publishing drawer in content only
- The new publishing drawer should not be instantiated in the common notes editor, it will be a common component between notes detail (3 dots option) and content editor only

hakermi:
> Some remarks:
> 
> * The FF should be in the content not notes in order to disable/enable the display of the new publishing drawer in content only
> * The new publishing drawer should not be instantiated in the common notes editor, it will be a common component between notes detail (3 dots option) and content editor only

Since the publishing drawer is a part of the common editor and depends of its components, i prefer to add it inside the editor even if not used in note creation mode, it still an option that is activated only for content.


for the FF i don't see any difference where i put it, keeping a temporary FF config in notes will not cause any problem.
i choose to put it in notes depends on the logic that is an editor option and a part of its behavior

hakermi:
> Some remarks:
> 
> * The FF should be in the content not notes in order to disable/enable the display of the new publishing drawer in content only
> * The new publishing drawer should not be instantiated in the common notes editor, it will be a common component between notes detail (3 dots option) and content editor only

https://github.com/Meeds-io/notes/pull/1152#issuecomment-2400509881

azayati (approved):
LGTM

PR #1157, opened 2024-10-11
fix: Fix image cropper styles in notes-editor - EXO-74717 - Meeds-io/MIPs#129

hakermi:
Fix image cropper styles in notes-editor


azayati (approved):
LGTM

PR #1171, opened 2024-10-16
fix: Fix import featured images - EXO-74718 - Meeds-io/meeds#2499

hakermi:
Prior to this change, when import notes with featured images, the featured image are not correctly imported due to wrong extraction of featuredId from exported fileName beside a wrong handle of different conflict actions. 
This PR makes sure to correctly extarct the export featured image id to correctly import featured images.


azayati (approved):
LGTM

PR #1249, opened 2024-12-11
fix: labels are not translated in publication drawer - EXO-75965 - meeds-io/MIPS#161

rdenarie:
Before this fix, the i18n file for noptespublicationDrawer is not loaded when the drawer is opened from the news detail. This commit add the load of the needed file on drawer creation

### DIFF
--- a/notes-webapp/src/main/webapp/vue-app/notes-publication/components/NotePublicationDrawer.vue
+++ b/notes-webapp/src/main/webapp/vue-app/notes-publication/components/NotePublicationDrawer.vue
@@ -298,6 +298,14 @@ export default {
       }
     }
   },
+  created() {
+    const lang = eXo.env.portal.language;
+    const urls = `/content/i18n/locale.portlet.notes.notesPortlet?lang=${lang}`;
+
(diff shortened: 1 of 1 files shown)


sofyenne (approved):

hakermi:
Hello,
Instead of this, why we don't just import notes i18n labels in newsdetails module ? which seems more organized and simpler, i don't think that this call is needed in notes context.

https://github.com/Meeds-io/content/blob/feature/experience/content-webapp/src/main/webapp/vue-app/news-details/main.js

as we did for the editor :
https://github.com/Meeds-io/content/blob/feature/experience/content-webapp/src/main/webapp/vue-app/news-activity-composer-app/main.js#L32

hakermi:
beside, all labels should be exposed in all the components of newsDetails even before open the publication drawer

https://github.com/Meeds-io/content/blob/feature/experience/content-webapp/src/main/webapp/vue-app/news-details/components/ExoNewsDetails.vue#L232

hakermi:
cc @rdenarie

PR #1273, opened 2025-01-07
fix: URL having Anchor element isn't displaying as expected - EXO-73493 - Meeds-io/meeds#2665.

akhanfir:
Before this change, on Activity stream, click on note URL having anchors, URL is cleaned (anchor element removed) and note is displayed. To resolve this problem, add the hash in the URL and scroll to the anchor. After this change, Url is persisted and the note displayed focusing on anchor.

### DIFF
--- a/notes-webapp/src/main/webapp/vue-app/notes/components/NotesOverview.vue
+++ b/notes-webapp/src/main/webapp/vue-app/notes/components/NotesOverview.vue
@@ -430,7 +430,19 @@ export default {
         this.stopGetStatus();
         this.exportStatus={};
       }
-    }
+    },
+    initialized() {
+      if (this.initialized) {
+        const urlHash = window.location.hash;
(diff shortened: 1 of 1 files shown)


rdenarie (changes requested):
2nd remark : anchor mechanisms are default mechanism. Why we must add a specific mechanism to apply this behaviour ?

If the anchor is present in the page, the browser must be able to find it.
We should not add something in vue for that

akhanfir:
@rdenarie You are correct; in the case of a normal click on a hash link or static content with hash links, the browser handles scrolling automatically. However, in our scenario, the content is dynamically mounted in the DOM after the hash link is accessed externally. As a result, the browser cannot find the target anchor section to scroll to. Therefore, we need to handle this case manually.

rdenarie (approved):

PR #1354, opened 2025-02-26
feat: Exclude SNV content type from content statistics measuring - MEED-8096 - Meeds-io/MIPs#162

hakermi:
Exclude SNV content type from content statistics measuring

### DIFF
--- a/notes-service/src/main/java/org/exoplatform/wiki/service/rest/NotesRestService.java
+++ b/notes-service/src/main/java/org/exoplatform/wiki/service/rest/NotesRestService.java
@@ -1442,13 +1442,13 @@ public class NotesRestService implements ResourceContainer {
   }
   
   @POST
-  @Path("/note/view/{noteId}/{lang}")
+  @Path("/note/view/{noteId}")
   @Operation(summary = "mark a note as viewed", method = "POST", description = "This marks a note as viewed.")
   @ApiResponses(value = { @ApiResponse(responseCode = "200", description = "Request fulfilled"),
       @ApiResponse(responseCode = "401", description = "User not authorized to get the note"),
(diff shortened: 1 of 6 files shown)


notes-service/src/main/java/org/exoplatform/wiki/service/rest/NotesRestService.java
```diff
@@ -1442,13 +1442,13 @@ public Response getFeaturedImageIllustration(@Context Request request,
...
   @POST
-  @Path("/note/view/{noteId}/{lang}")
+  @Path("/note/view/{noteId}")
   @Operation(summary = "mark a note as viewed", method = "POST", description = "This marks a note as viewed.")
   @ApiResponses(value = { @ApiResponse(responseCode = "200", description = "Request fulfilled"),
       @ApiResponse(responseCode = "401", description = "User not authorized to get the note"),
       @ApiResponse(responseCode = "500", description = "Internal server error") })
   public Response markNoteAsViewed(@Parameter(description = "News id") @PathParam("noteId") String noteId,
-                                   @Parameter(description = "News target lang") @PathParam("lang") String lang) {
+                                   @Parameter(description = "News target lang") @QueryParam("lang") String lang) {
```
rdenarie, line 1451: What is the reason of this change ?

rdenarie (changes requested):

notes-service/src/main/java/org/exoplatform/wiki/service/rest/NotesRestService.java
hakermi: because when lang is null in version language i will have incorrect non matching endpoint ends with ...id/

rdenarie (approved):

PR #1481, opened 2025-07-24
fix: Problem with hours selection when publishing - MEED-9451 - meeds-io/meeds#3492

akhanfir:
before this change, when request to publish, select Schedule and update the date from current day +1 to current day then update it again to choose another date, The hour is restricted to the one you can use for the current day. To fix this problem, reset minStartTime and minEndTime if the selected date is no longer today. After this change, if the selected day is not the current one, then it must be possible to select any hour.


Jihed525:
can you review this one @rdenarie ?

rdenarie (approved):

PR #1579, opened 2026-01-07
fix: Notes attachments are not saved in space drive folder - EXO-83542 - Meeds-io/meeds#3963

hakermi:
Note attachments were not saved in the space drive because the spaceId in the attachment app config was null. This update makes sure the config is populated with the correct spaceId.


rdenarie (changes requested):
Can you update the commit message to follow [developer guidelines](https://developer.exoplatform.org/#id-contributions) : 

- The meeds issue reference is missing
- The title must be what the commit does, not what the commit fixes
- the message must describe in detail what the commit does and must answer 2 questions :
  - why is this change needed ?
  - how does this change address the issue
Thanks

rdenarie (changes requested):
you change only the title and the text of the PR, not of the commit.

hakermi:
> you change only the title and the text of the PR, not of the commit.

I know, it will updated during merge

rdenarie (approved):

PR #1591, opened 2026-01-21
fix: Restore the summary and illustration note when reverting to a previous version- MEED-10085 - meeds-io/meeds#3943 

MayTekayaa:
Before this change, when restoring a note, the summary and featured image were not restored because these properties of each version of the note are saved, but they are not restored when the version details are retrieved. In addition, the properties of the first version of the note were not saved.
This commit adapts the code; change the pageHistory model and add  buildNotePageMetadata method when retrieving note versions to obtain the properties of each one with the appropriate version number.

### DIFF
--- a/notes-service/src/main/java/org/exoplatform/wiki/jpa/entity/PageVersionEntity.java
+++ b/notes-service/src/main/java/org/exoplatform/wiki/jpa/entity/PageVersionEntity.java
@@ -18,6 +18,7 @@
  */
 package org.exoplatform.wiki.jpa.entity;
 
+import io.meeds.notes.model.NotePageProperties;
 import jakarta.persistence.Column;
 import jakarta.persistence.Entity;
 import jakarta.persistence.GeneratedValue;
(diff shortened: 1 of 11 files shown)


rdenarie (changes requested):
Hi.
Can you complete your commit message ?
You perfectly describe WHAT is the problem, but 2 information are missing : 
WHY this problem occurs  ?
HOW your fix correct the problem ?

You can help by using
Before this fix, when restoring a note, summary and title where not restored because ...
This commit ensure to ....

Thanks

MayTekayaa:
> Hi. Can you complete your commit message ? You perfectly describe WHAT is the problem, but 2 information are missing : WHY this problem occurs ? HOW your fix correct the problem ?
> 
> You can help by using Before this fix, when restoring a note, summary and title where not restored because ... This commit ensure to ....
> 
> Thanks

Done!

rdenarie (approved):

PR #1657, opened 2026-04-06
feat: load publish extensions into publication drawer - EXO-85702 - exoplatform/eXIP#41

hakermi:
load publish extensions into publication drawer

### DIFF
--- a/notes-webapp/src/main/webapp/vue-app/notes-publication/components/NotePublicationDrawer.vue
+++ b/notes-webapp/src/main/webapp/vue-app/notes-publication/components/NotePublicationDrawer.vue
@@ -76,6 +76,14 @@
                           :has-featured-image="hasFeaturedImage"
                           :summary-max-length="summaryMaxLength"
                           @properties-updated="propertiesUpdated" />
+                        <template v-for="(component) in publishExtensions">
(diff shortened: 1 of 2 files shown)


notes-webapp/src/main/webapp/vue-app/notes-publication/components/NotePublicationDrawer.vue
```diff
@@ -180,6 +188,14 @@
                           }"
                           ref="advancedOption"
                           @update="updateAdvancedSettings" />
+                        <template v-for="(component) in publishExtensions">
```
boubaker: Is it better to use `<extension-registry-components` which will make the loop and listening on component modifications ?

hakermi: That will load extensions internally in that component while i need to load them outside to have control on the event save by using the execute function. thus i choose that way

boubaker: ok, I understand, for future extensibility then. Thanks for the explanation

boubaker (approved):

PR #1686, opened 2026-06-05
fix: Fix Notes editor application display issue in edit layout mode - EXO-87312

azayati:
Prior to this change, the Notes editor application is not well displayed centered in edit layout mode.
After this commit, some css changes are done in order to adjust the display position of the the Notes editor application.

### DIFF
--- a/notes-webapp/src/main/webapp/skin/less/notes/notes.less
+++ b/notes-webapp/src/main/webapp/skin/less/notes/notes.less
@@ -72,10 +72,8 @@
     }
   }
   #notesEditor {
-    position: fixed;
     height: 100%;
     top: 0;
-    z-index: 1031;
     background: #eeeeee;
 
     .notes-content-wrapper {
@@ -83,7 +81,7 @@
     }
 
     .notes-content {
-      height: ~"calc(var(--100vh, 100vh) - 100px)";
+      height: ~"calc(var(--100vh, 100vh) - 136px)";
       overflow: auto;
       z-index: 5;
       background-color: #eeeeee;


notes-webapp/src/main/webapp/skin/less/notes/notes.less
```diff
@@ -43,6 +43,23 @@
...
 
+.contentEditorContainer {
+  .layout-section-content {
+    width: 100% !important;
+    padding-left: 0 !important;
+    padding-right: 0 !important;
+  }
+}
+
+#layoutEditor {
```
boubaker: May be avoid adding specific cases for portlets, WDYT ?

azayati: The basic definition of  .notesEditor.#notesEditor is:
    position: fixed;
    height: 100%;
    top: 0;
    z-index: 1031;
    background: #eeeeee;
I prefered to avoid changing it which will impact the display behavior of view mode, and just override it in the case of edit mode for #layoutEditor.notesEditor.#notesEditor

boubaker: sure, I understand, in fact I would like to look at this as a generic concern which may be addressed globally for any full page app.

azayati: ok I let u see if a modification we can discuss is needed, but I think we have this problem only with notes-editor and news-editor for full page

boubaker (approved):

PR #1699, opened 2026-06-23
Fix knowledge base auto-import: subspace template and deletion order

houssemexo26:
### Fixes

1. **Filter out subspace templates when creating knowledge base spaces**
   - Knowledge base spaces were failing to create when the selected community template was a subspace template (template 3). Added filtering via `getSubspaceTemplateIds()` to ensure only top-level templates are used.

2. **Prevent EntityNotFoundException noise during notes import**
   - Reversed note deletion order so children are deleted before parents, preventing cascade-related EntityNotFoundException.
   - Downgraded `deleteNote`'s pre-throw log from ERROR to WARN since the exception is the proper error signal and callers handle it.

### DIFF
--- a/notes-service/src/main/java/org/exoplatform/wiki/service/NotesAutoImportService.java
+++ b/notes-service/src/main/java/org/exoplatform/wiki/service/NotesAutoImportService.java
@@ -296,7 +296,12 @@ public class NotesAutoImportService implements Startable {
     space.setPrettyName(prettyName);
     space.setVisibility(Space.HIDDEN);
     space.setRegistration(Space.CLOSED);
-    Optional<SpaceTemplate> communitySpaceTemplate = spaceTemplateService.getSpaceTemplates().stream().filter(template -> template.getLayout().equals("community")).findFirst();
+    List<Long> subspaceTemplateIds = spaceTemplateService.getSubspaceTemplateIds(superUserIdentity.getUserId());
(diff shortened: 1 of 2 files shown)


notes-service/src/main/java/org/exoplatform/wiki/service/NotesAutoImportService.java
```diff
@@ -296,7 +296,12 @@ private static Space createSpace(String prettyName,
     space.setPrettyName(prettyName);
     space.setVisibility(Space.HIDDEN);
     space.setRegistration(Space.CLOSED);
-    Optional<SpaceTemplate> communitySpaceTemplate = spaceTemplateService.getSpaceTemplates().stream().filter(template -> template.getLayout().equals("community")).findFirst();
+    List<Long> subspaceTemplateIds = spaceTemplateService.getSubspaceTemplateIds(superUserIdentity.getUserId());
+    Optional<SpaceTemplate> communitySpaceTemplate = spaceTemplateService.getSpaceTemplates()
+                                                                         .stream()
+                                                                         .filter(template -> template.getLayout().equals("community"))
```
boubaker, line 302: Sorry, we shouldn't embed a specific space template name, it should be done switch a specific characteristic instead. Can you please open a feedback about it to address it by product team?

houssemexo26: I'll drop this fix and let you take it from here since the right approach needs a product decision on what characteristic to use. Thanks for the feedback!"

PR #1706, opened 2026-06-25
fix: Fix duplicated note blank page - EXO-87922

azayati:
Prior to this change, when duplicating a note, a blank page is displayed. This is caused by the fact that "canRedact" prop value is computed from NotesEditorDashboard component then passed to NoteFullRichEditor component only when the url notes-editor url contains "parentNoteId" param which is not the case for duplicated note url like "notes-editor?noteId=XX&translation=&spaceGroupId=XXXXX"
After this commit, we will ensure to compute "canRedact" prop value even without "parentNoteId" url param which covers also the duplicated note case.

### DIFF
--- a/notes-webapp/src/main/webapp/vue-app/notes-editor/components/NotesEditorDashboard.vue
+++ b/notes-webapp/src/main/webapp/vue-app/notes-editor/components/NotesEditorDashboard.vue
@@ -242,10 +242,10 @@ export default {
       this.parentPageId = parentNoteId === 'null' ? null : parentNoteId;
       this.spaceId = urlParams.get('spaceId');
       this.spaceGroupId  = urlParams.get('spaceGroupId');
-      this.getTargetSpaceId(this.spaceGroupId);
       this.spaceDisplayName  = urlParams.get('spaceName');
       this.note.parentPageId = this.parentPageId;
     }
+    this.retrieveSpaceInformation(this.spaceGroupId);
     if (urlParams.has('owner')) {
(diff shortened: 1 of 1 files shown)


notes-webapp/src/main/webapp/vue-app/notes-editor/components/NotesEditorDashboard.vue
```diff
@@ -242,10 +242,10 @@ export default {
       this.parentPageId = parentNoteId === 'null' ? null : parentNoteId;
       this.spaceId = urlParams.get('spaceId');
       this.spaceGroupId  = urlParams.get('spaceGroupId');
-      this.getTargetSpaceId(this.spaceGroupId);
       this.spaceDisplayName  = urlParams.get('spaceName');
       this.note.parentPageId = this.parentPageId;
     }
+    this.getTargetSpaceId(this.spaceGroupId);
```
boubaker: it's not a `get`, is it a `compute` or `retrieve` ?

azayati: yes retrieves targetSpaceId and canRedact, we can maybe put retrieveSpaceInformation, WDYT ?

boubaker (approved):

notes-webapp/src/main/webapp/vue-app/notes-editor/components/NotesEditorDashboard.vue
boubaker: yes cool

PR #1710, opened 2026-07-05
feat: Add Notes MCP tools — versions, cover, summary & translations - EXO-88352

bmestrallet:
## What

Extends the Notes add-on's existing `NoteMcpTool` so EVA can manage a note's **history and metadata**, not just its title/content.

| Tool | Kind |
|---|---|
| `get_note_versions` | read-only — version number, title, content, date, author |
| `restore_note_version` | approval — roll back to a `version_number` |
| `set_note_cover` / `remove_note_cover` | approval — cover ("featured") image from URL / base64 / ACL-checked attachment |
| `set_note_summary` | approval — the summary, previously write-once on create |

Also surfaces `summary` in `get_note` (it was set on create but never returned or editable).

## Two note-cover bugs fixed

- **"Illustration never applied"** — the cover must go through `NoteService.saveNoteMetadata` (writes the file **and** the `featuredImageId` metadata). `saveNoteFeaturedImage` alone only stores the file, so nothing references it.
- **Cover flashed then disappeared** — after the metadata write, propagate onto a new version via `createVersionOfNote`, exactly like the native update flow. Otherwise the cover lives only on the page metadata and the front-end (which resolves it through the published version) shows it briefly then drops it.

## Reuse
`set_note_cover` reuses `UploadToolUtils.resolveImage` (SSRF-guarded fetch + ACL-enforced attachment read) → `materialize` → `uploadId`. All write tools are approval-gated and act as the current user with the note's ACL enforced.

## ⚠️ Stacked on Meeds-io/mcp-server#13
This depends on `UploadToolUtils`, which is new in **#13**. `addon.meeds.mcp-server.version` is bumped to `7.3.x-ai-contribution-SNAPSHOT` (aligning it with the other addons on this branch). **CI will be red until #13 merges and publishes that snapshot** — please review #13 first.

## Tests & verification
- 24 module tests green (versions, restore, cover, summary, guards).
- Deployed + verified live with EVA: browsed versions, set covers (render fix confirmed on-screen after the version-propagation fix), set/read summaries.

Task: EXO-88352 · Next Notes PR: translations discovery (list a note's languages + read per-language).

🤖 Generated with [Claude Code](https://claude.com/claude-code)


bmestrallet:
Folded the **translations** work into this PR (2nd commit):

- `get_note_translations(note_id)` — list a note's translation languages (`NoteService.getPageAvailableTranslationLanguages`).
- `get_note` gains an optional `language` — read a note in a specific translation.
- **Fix:** `update_note` with a `language` was overwriting the **default** note (it applied the translated title/content then called `updateNote`, which writes the default page). Reordered to save the default first and carry the translation only into the language version via `createVersionOfNote`, matching the native flow.

Deliberately **out of scope** (tracked as a follow-up): making a new translation inherit the default note's cover, and a language-targeted `set_note_cover`. eXo stores covers **per language** across page/version metadata; that needs the native metadata write traced first rather than guessed.

Tests: 26 module tests green. Still stacked on Meeds-io/mcp-server#13.

notes-service/src/main/java/io/meeds/notes/mcp/NoteMcpTool.java
```diff
@@ -261,6 +307,164 @@ public List<NoteModel> searchNotes(String query,
                   .toList();
   }
 
+  // Lists the version history of a note: each version's number, id, title,
+  // content, date and author. Read as the current user, so the note's ACL is
+  // enforced. Read-only.
+  public List<NoteVersionModel> getNoteVersions(long noteId, String language) throws IllegalAccessException,
```
ahamdi: Could you ask IA to add Java code comments so that this gets included in the javadocs.

bmestrallet: On very new method or just this one?

bmestrallet: Done

bmestrallet:
@ahamdi Done — converted the `NoteMcpTool` method comments to proper `/** */` Javadoc so they're included in the generated Javadocs. Pushed.

boubaker:
## Review: Notes MCP tools — versions, cover, summary & translations (stacked on mcp-server#13)

### Overview
Extends `NoteMcpTool` with version history/restore, cover (featured image) set/remove, summary set, and translation discovery. Also fixes two described cover bugs (wrong write path, missing version-propagation) and exposes `summary` in `get_note`. I traced the actual `NoteServiceImpl`/`NoteDataStorage` implementations behind these calls (not just the MCP layer) to verify the described fixes and check for new regressions.

---

### 🔴 Critical — `remove_note_cover` doesn't durably remove the cover; it resurrects a reference to the just-deleted file

`removeNoteCover` does, in order:
```java
noteService.removeNoteFeaturedImage(noteId, existing.getId(), lang, false, currentUserIdentityId(username)); // deletes the file + clears FEATURED_IMAGE_ID from metadata
noteService.createVersionOfNote(note, username, true); // <-- `note` is the ORIGINAL object, its .getProperties().getFeaturedImage() still holds the OLD id
```
I traced `NoteServiceImpl.createVersionOfNote` (line ~1020-1057): when `note.getLang() != null` and `note.getProperties() != null` (both true here — nothing cleared `note`'s in-memory properties after the removal), it calls `saveNoteMetadata(note.getProperties(), note.getLang(), ...)` again. Inside `saveNoteMetadata`, since the stale `featuredImage` object is non-null and not marked `isToDelete()`, it falls into `saveNoteFeaturedImage(note, featuredImage)`, which — since `featuredImage.getUploadId()` is null (it's a "read back" id-only object, not a fresh upload) — skips the file-write branch and just returns the existing id unchanged. Back in `saveNoteMetadata`, that non-null id causes `properties.put(FEATURED_IMAGE_ID, ...)` to **re-add the id `removeNoteFeaturedImage` had just removed**, and persists it via `metadataService.updateMetadataItem`.

Net effect: `removeNoteFeaturedImage` correctly deletes the file (`fileService.deleteFile(id)`) and clears the metadata — and then, one line later, `createVersionOfNote` writes that same (now-deleted) file id straight back into the metadata. The note ends up with a cover reference pointing at a file that no longer exists (broken image), and functionally the "remove" never took effect from the metadata's perspective.

**Fix**: clear the local reference before creating the version, e.g. `properties.setFeaturedImage(null); note.setProperties(properties);` right after the `removeNoteFeaturedImage` call, so `createVersionOfNote`'s internal resave sees a null `featuredImage` (which, per `saveNoteMetadata`'s own logic, means "don't touch/re-add the id" rather than "clear it" — but since the id was already cleared by `removeNoteFeaturedImage`, leaving it alone is exactly correct here).

---

### 🟠 High — the `language` parameter isn't honored by the `Page` object used for the actual write in several tools

`setNoteCover`, `setNoteSummary`, `removeNoteCover`, and `restoreNoteVersion` all start with:
```java
Page note = getNoteById(noteId);   // single-arg overload → resolves via the CALLER's own session locale, not the `language` param
```
...and only later derive a `lang` string (`StringUtils.isBlank(language) ? note.getLang() : language`) for the `saveNoteMetadata`/`removeNoteFeaturedImage` calls. But `noteService.createVersionOfNote(note, ...)` and `noteService.restoreVersionOfNote(versionName, note, ...)` key the *version's own language* off `note.getLang()` (confirmed in `NoteServiceImpl.createVersionOfNote`: `if (note.getLang() != null) { ...saveNoteMetadata(properties, note.getLang(), ...) }`, and `addPageVersion` stamps `pageVersionEntity.setLang(page.getLang())`) — not off the `lang` local variable, and not off the tool's `language` argument.

So: if a caller's own session locale is `en` but they explicitly pass `language: "fr"` to `set_note_cover`/`set_note_summary`/`remove_note_cover`/`restore_note_version` (exactly the use case these tools advertise — "Optional note language/translation to target"), the metadata write for the *properties* correctly targets `"fr"`, but the **new version created afterward is tagged/created under `"en"`** (the viewer's own locale) instead of `"fr"` — a mismatch between where the property change is recorded and where the version history says it happened. `get_note_versions(note_id, language="fr")` afterward would not show this new version at all.

`getNoteVersions`/`versionsHistory` are unaffected — they pass the explicit `language` straight through to `getVersionsHistoryOfNoteByLang`, ignoring `note.getLang()`.

**Fix**: fetch `Page note = getNoteById(noteId, language);` (the two-arg overload already added in this PR) in these four methods, so `note.getLang()` reflects the requested target language throughout.

---

### 🟡 Medium — `update_note` on a translation discards that translation's own summary/cover

In the reworked `updateNote` translation branch:
```java
Page defaultNote = noteService.getNoteByIdAndLang(noteId, currentUserAclIdentity, null, null); // lang=null = the default note
if (defaultNote != null && defaultNote.getProperties() != null) {
  note.setProperties(defaultNote.getProperties());   // always the DEFAULT note's properties
}
```
I confirmed translations carry their own distinct `NotePageProperties` (summary/cover), separate from the default note's. This unconditionally overwrites whatever summary/cover the *translation itself* had (e.g. previously set via `set_note_summary`/`set_note_cover` with `language: "fr"`) with the *default* note's summary/cover, every time `update_note` is called with a `language`. The comment justifies this for the "translation renders with the default's cover" case (first-ever translation, no properties of its own yet), but it's applied unconditionally, clobbering a translation that already established its own metadata.

**Fix**: only fall back to the default note's properties when the translation doesn't already have its own (`noteService.getNoteByIdAndLang(noteId, identity, null, language)` first; only inherit from `defaultNote` if that comes back null/propertyless).

---

### Confirmed as correctly fixed
- The core "cover never applied" fix (routing through `saveNoteMetadata` instead of the file-only `saveNoteFeaturedImage`) and the "flashed then disappeared" fix (adding the `createVersionOfNote` call after metadata writes in `setNoteCover`/`setNoteSummary`) are both real and match the exact idiom used by the native `NotesRestService.updateNoteById` REST path for translations (persist unmutated, then mutate + version) — good corroboration these follow established conventions rather than inventing a new pattern.

### Minor / style
- `NotePage.vue`'s version-switch handler now merges `featuredImage`/`summary` from the version's properties with a fallback to the note's current properties — reasonable client-side defensive patch, no issue found.
- `pom.xml`'s `addon.meeds.mcp-server.version` bump is called out explicitly as breaking CI until mcp-server#13 merges — good, transparent about the stacking.

### Test coverage
Good breadth (24 tests: versions, restore, cover, summary, guards, translations). None of them exercise the two bugs above, because both require state that differs across calls in ways the mocks don't model:
- No test asserts that after `remove_note_cover`, a *subsequent* read (`get_note`/`get_note_versions`) still shows no cover — the existing test only verifies `removeNoteFeaturedImage` was called with the right args, not that the following `createVersionOfNote` doesn't undo it.
- No test passes a `language` different from the mocked current-user locale (`"en"` throughout) to `set_note_cover`/`set_note_summary`/`remove_note_cover`/`restore_note_version`, so the `note.getLang()` vs. requested `language` mismatch never surfaces.

---
🤖 This review was generated by [Claude Code](https://claude.com/claude-code).

bmestrallet:
@boubaker — thanks for the deep trace, all three landed:

- **🟡 Medium** (translation clobber) — already fixed in the stacked **#1713** via `resolveBaseProperties`: the write is based on the target language’s own current properties, inheriting the default only for a brand-new translation.
- **🔴 Critical** (`remove_note_cover` resurrecting the just-deleted id) — fixed: after `removeNoteFeaturedImage`, we now null the featured image on the in-memory `note` before `createVersionOfNote`, so the version resave sees null (= “leave the stored id alone”) instead of re-adding the deleted file id. Exactly the idiom `setNoteSummary` already relied on.
- **🟠 High** (language not honored for the version write) — fixed: `setNoteCover`, `setNoteSummary`, `remove_note_cover` and `restore_note_version` now load via the two-arg `getNoteById(noteId, language)`, so `note.getLang()` = the requested language and `createVersionOfNote`/`restoreVersionOfNote` tag the version under it. Confirmed the two-arg overload resolves blank language identically to the single-arg, so the default path is unchanged.
- **Tests** — added the two gaps you flagged: `removeNoteCoverShouldNotResurrectRemovedCover` and language≠locale version-tagging (`set_note_cover` + `restore_note_version` via `ArgumentCaptor`). Each fails against the pre-fix code; full run 32 green.

All in **#1713** (commit 447905f5a, stacked on this PR), since it already owns these methods. This PR (#1710) is unchanged.

PR #1722, opened 2026-07-09
fix: redirection to page not found when opening note from snack bar -  EXO-88375

mkrout:
Prior to this, clicking "View" or "Copy link" on a note could redirect to a "page not found" error whenever the Notes application had been added to a navigation node other than the default one - for example a node manually added from the site navigation editor to expose a user's personal notes on a "people" page.

This happened because the note's URL was always built with the default node segment hardcoded ("notes" for spaces in Utils#getPageUrl and NotePermanentLinkPlugin), instead of the navigation node actually hosting the application. For personal notes specifically, the frontend also unconditionally rewrote the note URL to "dashboard/notes/{id}", discarding the node the user was actually browsing when the notes editor popup was opened.

This commit fixes the problem by resolving, server-side, the navigation node that hosts the Notes application for a given space (falling back to the historical "notes" node when it exists, to preserve current behavior), and by passing the current node URI through as a "notePageUri" parameter when opening the notes editor popup, so personal notes build their "View"/copy-link URL from the node the user was actually on instead of the hardcoded "dashboard/notes".

### DIFF
--- a/notes-webapp/src/main/webapp/vue-app/notes-editor/components/NotesEditorDashboard.vue
+++ b/notes-webapp/src/main/webapp/vue-app/notes-editor/components/NotesEditorDashboard.vue
@@ -644,7 +644,6 @@ export default {
       return draftNote;
     },
     redirectAfterSaveLink(note) {
-      console.log('redirectAfterSaveLink', note);
       if (this.webPageUrl) {
         return this.webPageUrl;
       } else {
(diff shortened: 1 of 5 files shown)


notes-webapp/src/main/webapp/vue-app/notes-editor/components/NotesEditorDashboard.vue
```diff
@@ -641,6 +644,7 @@ export default {
       return draftNote;
     },
     redirectAfterSaveLink(note) {
+      console.log('redirectAfterSaveLink', note);
```
ahamdi: To be removed please

ahamdi (changes requested):

ahamdi (approved):

PR #1728, opened 2026-07-17
Merge exip7.3.0.1

AzmiTouil:


boubaker (commented):
## Review: Content Categorization (eXIP — Tribe activity 473782 / news 46633)

Cross-checked this PR against the companion PRs (`social#5889`, `content#832`) and the eXIP spec.

### 1. Contributes to the "browse content by category" drawer missing posted Notes
`NotePage.vue`'s `updateCategories()` correctly falls back to a direct link when there's no activity yet:
```js
const objectType = this.note.activityId ? 'activity' : 'notes';
const objectId = this.note.activityId ? this.note.activityId : this.note.id;
```
This is the *right* behavior for the "manage categories" action itself (nice catch handling the not-yet-posted case, unlike the `content` PR's News equivalent). But it means that once a note **is** posted, its category link lives under `CategoryObject("activity", activityId, ...)`, not `"notes"`. The new `GET /categories/{id}/entries?types=news,notes` endpoint added in `social#5889` filters strictly on `objectType IN ('news','notes')`, so posted notes (the common case) won't show up when browsing their own category — see the `social#5889` review comment for the full explanation. Worth a cross-repo fix: either the browse endpoint needs to also resolve through the activity link, or (better, and closer to what the spec's "Software Architecture" section describes) the direct note-level link should be the source of truth with a listener syncing the activity link, rather than the other way around.

### 2. Entry preview never carries its own category badge
`NoteCategoryPlugin.getEntryItem()` builds the `CategoryEntryItem` from `noteService.getNoteById(objectId)`, but `Page.categoryIds` is only ever populated by `NotesRestService.populateCategories()` at the REST layer — not inside the service call used here. So `CategoryEntryItem.categoryIds` is always empty, and the per-item category chip in `EntryListItem.vue` (social) never renders for note entries. Minor/cosmetic, same issue as the News side in `content#832`.

---
Reviewed by Claude

boubaker (commented):
## Follow-up review

I don't see any new commits here since my last review (same commit list/diff). Restating what's still open on this side:

- The minor cosmetic gap still applies: `NoteCategoryPlugin.getEntryItem()` builds `CategoryEntryItem` from `noteService.getNoteById()`, which doesn't populate `Page.categoryIds` (only `NotesRestService.populateCategories()` does, at the REST layer) — so the entry preview's category badge won't render for note entries in the browse drawer. `content#832` just fixed the equivalent gap on the News side via a `getCategoryIds(news)` fallback inside `getEntryItem()`; the same fallback (`NoteCategoryPlugin.getCategoryIds(note)` is already a static method here) would close this for notes too.
- The cross-repo browse-by-category gap also still applies here: once a note is posted, its category link lives under `CategoryObject("activity", activityId, ...)` (this repo's own `NotePage.vue` `updateCategories()` — which was already doing the right thing, unchanged), but the browse drawer is opened with `['news', 'notes']` only, so posted notes won't show up when browsing by category. See my comment on `social#5889` for where I think this needs to be resolved.

Not approving yet — nothing to re-review since the last pass, and the open items above still stand.

---
Reviewed by Claude

boubaker (commented):
## Follow-up review (full fresh re-pass against current HEAD)

🔴 **Critical (shared with social#5889)** — same browse-drawer gap
[`NotePage.vue:939`](https://github.com/Meeds-io/notes/blob/88383cad46003a105636368572f2f2124151b810/notes-webapp/src/main/webapp/vue-app/notes/components/NotePage.vue#L939)
```js
this.$refs.entryListDrawer.open(category.id, ['news', 'notes']);
```
Same issue as content's equivalent call site — once a note is posted, its category link is `CategoryObject("activity", activityId, ...)`, not `"notes"`, so this drawer (opened from the note's own category chip) won't find the note whose chip was just clicked. See the `social#5889` comment for the proposed fix.

**Fix:** depends on the social-side fix; likely no client change needed if resolved server-side.

---

🟡 **Medium** — zero test coverage for any of the new category logic
No test files are touched in this PR at all — `NoteCategoryPlugin`, the direct-link fallback for unposted notes, and `updateCategories()` in `NotePage.vue` have no coverage.

---

**Correction to my own earlier review:** I previously flagged `NoteCategoryPlugin.getEntryItem()` as having the same missing-`categoryIds` cosmetic gap as News originally had. Re-reading the actual source at this exact commit ([`NoteCategoryPlugin.java:154`](https://github.com/Meeds-io/notes/blob/88383cad46003a105636368572f2f2124151b810/notes-service/src/main/java/io/meeds/notes/plugin/NoteCategoryPlugin.java#L154)), it calls the static `getCategoryIds(note)` helper (which resolves via `CategoryLinkService`), not the unpopulated `note.getCategoryIds()` instance getter I'd assumed — that was a misread on my part in the prior pass. Retracting that finding for notes; it was already correct, ahead of content's original bug on the News side.

What's good: `toCategoryObject()`/`updateCategories()` correctly handled the unposted-note direct-link fallback from the start — this is the pattern content later copied to fix its own equivalent bug.

---
Reviewed by Claude

boubaker (approved):
## Final review — all findings resolved

Re-verified against the latest commit. The new `NoteCategoryPluginTest.java` closes the test-coverage gap I'd flagged, covering `toCategoryObject()`'s posted/unposted resolution and confirming `getEntryItem()` resolves categoryIds via the static `getCategoryIds(note)` helper (not a stale instance getter) — matching what I'd retracted as a misread in my previous pass.

The browse-drawer call site (`NotePage.vue`) is unchanged (`['news', 'notes']`), which is correct now that resolution happens transparently server-side in `social#5889`.

No new issues found. Approving.

---
Reviewed by Claude

